Issue #25, opened 2016-03-07
Facebook Video/Post embeds not working. 


rinatkhaziev:
Which is ironic :) 

I worked this around with following code -

```
static function wire_fb_embed_handlers() {
	wp_embed_register_handler( 'facebook-ia-post', JETPACK_FACEBOOK_EMBED_REGEX, [ __CLASS__, 'ia_fb_embed_handler' ] , 0 );
	wp_embed_register_handler( 'facebook-ia-video', JETPACK_FACEBOOK_VIDEO_ALTERNATE_EMBED_REGEX, [ __CLASS__, 'ia_fb_embed_handler' ] , 0 );
}

static function ia_fb_embed_handler( $matches, $attr, $url ) {
	$class = wp_in( '/posts/', $url ) ? 'post' : 'video';
	return sprintf( '<div class="fb-%s" data-allowfullscreen="true" data-href="%s" data-width="%d"></div>', esc_attr( $class ), esc_url( $url ), (int) Helper::get_content_width( 'ia' ) );
}
```

Obviously, this needs to be cleaned up since I have all kinds of helpers which are probably not needed + a check to see if Jetpack is activated

philipjohn:
Thanks @rinatkhaziev - do you have an example to point us at? :)

rinatkhaziev:
@philipjohn I just updated the original comment, it covers links like https://www.facebook.com/rinat.khaziev/posts/10208586784288538 and https://www.facebook.com/coslive/videos/10153831160106648/

philipjohn:
@rinatkhaziev How does the above fix work out for you?

rinatkhaziev:
@philipjohn works like a charm! 

Closing this one.

rinatkhaziev:
@philipjohn when are you planning to merge this into wp.com?

PR #39, opened 2016-03-08
Jetpack compat: YouTube and Facebook embeds

### LINKED ISSUE #20: Jetpack YouTube wrappers breaks feed
Following on from #17 we've discovered that Jetpack's filters for YouTube embeds is causing the DOM parsing to fail to turn YouTube embeds into the right FBIA format.

Jetpack adds a `span` with a class of `embed-youtube` and that simple addition confuses this plugin which can no longer do the transformation.

I'm thinking we have 3 possible solutions:

1. Make our DOM transformation more liberal/forgiving of additional HTML to account for this
1. Have Jetpack stop adding it's wrapping `span` in feeds in general, assuming it does so for front-end purposes.
1. Have our plugin remove all filters on `the_content` before [grabbing the content](https://github.com/Automattic/facebook-instant-articles-wp/blob/bed2ee8438ec937eafaafafb2376d4c827fc0bb3/class-instant-articles-post.php#L206), circumventing oEmbed entirely. It would no longer rely on DOM manipulation.

The first option would be great, but could be complicated. I'm talking to Jetpack about the second option. The third option is likely to be too big a change to the fundamental way the plugin works.

bjornjohansen:
Do not "fix" bare YouTube URLs on the own line as we have have oEmbed to handle those.

By unregistering the embed handler that Jetpack registers, we avoid the Jetpack YouTube HTML output altogether and let the oEmbed handler handle it.

Fixes #20

### DIFF
--- a/compat.php
+++ b/compat.php
@@ -21,3 +21,10 @@ if ( defined( 'GAWP_VERSION' ) && ! defined( 'GAWP_IA_COMPAT' ) ) {
 	$gawp->init();
 }
 
+// Load support for Jetpack
+if ( defined( 'JETPACK__VERSION' ) ) {
+	include( dirname( __FILE__ ) . '/compat/class-instant-articles-jetpack.php' );
+	$jp = new Instant_Articles_Jetpack;
+	$jp->init();
+}
+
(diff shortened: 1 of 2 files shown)


bjornjohansen:
c422595 takes care of proper Facebook post embeds.
Fixes #25

PR #40, opened 2016-03-08
Fix notices when registering DOM transformation filters with new priorities

bjornjohansen:
Check if the stack key exists before checking if the entry is an array (to avoid notices) when registering DOM transformation filters


philipjohn:
@bjornjohansen Would you be able to rebase this PR please so we can cleanly merge it in please?

everton-rosario:
@bjornjohansen Is this still possible to be rebased?

@philipjohn, if @bjornjohansen don't have the free cycles to rebase this, will you able to do a new PR and then we can get it reviewed and merged?

diegoquinteiro:
We don't use these transformations filters anymore since v2+.

PR #41, opened 2016-03-08
Add support for subtitles through the filter instant_articles_subtitle

### LINKED ISSUE #21: Add support for subtitles
The [format includes a subtitle](https://developers.facebook.com/docs/instant-articles/reference/cover), using a H2. We should probably add support for this, probably using a hook, to allow folks to add a subtitle when desired.

bjornjohansen:
Fixes #21

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -13,6 +13,9 @@ class Instant_Articles_Post {
 	/** @var string Cached version of the Instant Article body */
 	protected $_content = null;
 
+	/** @var string The post’s optional subtitle */
+	protected $_subtitle = null;
+
 	/**
 	 * Setup data and build the content
 	 *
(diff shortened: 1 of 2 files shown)


dannydehaan:
You can add an action on "instant_articles_article_head". Like this:

<?php
add_action( 'instant_articles_article_head', 'action__instant_articles_article_head' );

function action__instant_articles_article_head() {
	echo '<subtitle>Your subtitle here.</subtitle>';
}
?>

bjornjohansen:
@dannydehaan: Almost :-) Your logic is correct, but the correct implementation would be something in the lines of:
```
<?php
add_action( 'instant_articles_article_header', 'foobar' );

function foobar() {
	echo '<h2>Your subtitle here</h2>';
}
```
Notice the action name is different, as the subtitle should be in the `<article><header>` section, not in `<head>`.A bit confusing yes, but I didn’t create the spec :-) 
Also the `<h2>` element must be outputted.

The PR adds logic that handles this, even if the spec should change (which I’ve seen it do several times), so that plugins/themes that want to add a subtitle doesn’t have to worry about the spec at all.

dannydehaan:
@bjornjohansen, you're absolutely right! I'm working on instant articles for a client. I thought i had did it right, but i got some feedback from Facebook with the message the <subtitle> element was illegal. Now i've created a custom template.php and filtered the default template with ''instant_articles_render_post_template'. That did work for now :-).

amitmalewar:
Hello guys, 

Help me to get a subtitle in Facebook Instant articles. I am using Newspaper WP theme.

PR #42, opened 2016-03-09
Small refactor on the tests

brunohulk:
- Fix wrong indentation.
- Changing the methods to use the same pattern for the open brackets.
- Removing unnecessary spaces.


philipjohn:
Thanks for this @brunohulk !

We prefer to stick to the [WP coding standards on brace style](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style) but the whitespace changes would be welcome.

Do you want to tweak this PR accordingly, and rebase so we can get it merged?

Thanks!

everton-rosario:
If we get no response before releasing 3.0, will you be able to rebase it @philipjohn ?

PR #43, opened 2016-03-09
Support for related articles (and YARPP)

bjornjohansen:
Add support for suggesting related articles to FB.

Includes a compat file for Yet Another Related Posts Plugin (YARPP)

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -659,6 +659,68 @@ class Instant_Articles_Post {
 		return $article_style;
 	}
 
+	/**
+	 * Get whether we shall output a list of related articles in the footer
+	 *
+	 * @since 0.2
+	 * @return bool Whether we shall output a list of related articles in the footer.
+	 */
+	function use_related_articles_in_footer() {
+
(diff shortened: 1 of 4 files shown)


class-instant-articles-post.php
```diff
@@ -659,6 +659,68 @@ function get_article_style() {
...
+	 *         @type string  $url           The URL to the related article
+	 *         @type bool    $is_sponsored  Whether this is a sponsored link or not
+	 *     }
+	 * }
+	 */
+	function get_related_articles() {
+
+		$related_articles = array();
+
+		/**
```
philipjohn: Looks like the spacing got messed up here

philipjohn:
There's a bit of spacing to fix, and a rebase needed then I think we should be okay to merge this

diegoquinteiro:
This code is outdated and does not affect v2+ at all. We are not using `template.php` for anything anymore - it should be removed.

Markup is composed using SDK Elements on the Instant_Articles_Post (method `to_instant_article()`)

PR #49, opened 2016-03-10
make fetching of custom post types possible

dazwiafl:
added possibility for changing post_type of fetched posts for feeds (needed it for custom post types ;) )

```
//declaration in functions.php
add_action( 'init', 'set_instant_articles_constants', 9);
function set_instant_articles_constants(){
  define( 'INSTANT_ARTICLES_POST_TYPE', 'desired_post_type_1' );
  /* OR */
  define( 'INSTANT_ARTICLES_POST_TYPE', serialize( array('desired_post_type_1', 'desired_post_type_2' ) ) );
}
```

### DIFF
--- a/facebook-instant-articles.php
+++ b/facebook-instant-articles.php
@@ -140,6 +140,15 @@ function instant_articles_query( $query ) {
 		$query->set( 'posts_per_page', 100 );
 		$query->set( 'posts_per_rss', 100 );
 
+		// Let users define their own feed post_type
+		if( defined( 'INSTANT_ARTICLES_POST_TYPE' ) && INSTANT_ARTICLES_POST_TYPE){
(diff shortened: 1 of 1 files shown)


bjornjohansen:
There is already other ways of adding CPTs to feeds. I’m not convinced this is a feature we should add in the plugin. I’m open to adding a filter to achieve this, though.

A look at [core ticket 12943](https://core.trac.wordpress.org/ticket/12943) should give you an idea of at least one way of doing this.

You should also be able to use the `pre_get_posts` filter to [modify the query like we already do in the plugin](https://github.com/Automattic/facebook-instant-articles-wp/blob/master/facebook-instant-articles.php#L137).

dazwiafl:
thx @bjornjohansen 
i'll have a look in this. thx :)

PR #52, opened 2016-03-10
oEmbed tweaks and improvements

rinatkhaziev:
Remove a bunch of if/elseif checks to determine provider name, instead run $providerURL through a preg_match check and then extract service name from domain.

Check whether oembed result is iframe, and if it's not, wrap it in iframe, and then wrap in figure.op-social

I also removed [if ( $provider_name ( check ](https://github.com/Automattic/facebook-instant-articles-wp/compare/master...rinatkhaziev:master#diff-2555b2b5910c9a6a7cad07726635a883L50) because if it's present, and somehow $provider_name was not matched properly instant_articles_embed_get_html() won't be called, resulting in regular embed code (and some of them are not wrapped in iframe) and breaking the feed.

### DIFF
--- a/embeds.php
+++ b/embeds.php
@@ -1,5 +1,4 @@
 <?php
-
 /**
  * Oembed/embed handling
  *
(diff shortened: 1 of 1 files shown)


bjornjohansen:
The specific provider checks are there for a reason. Please consult the reference: https://developers.facebook.com/docs/instant-articles/reference/social

Also it doesn’t matter if the oembed result comes with an iframe or not.

rinatkhaziev:
Here's a screenshot of the error in previously approved feed - 
screenshot 2016-03-10 14 07 35

PR #59, opened 2016-03-14
add vimeo.com into oEmbed $provider_name

abdusfauzi:
Fix vimeo.com oEmbed figure wrapping.


kkranzo:
When will just a basic iframe (wrapping any standard iframe/non oEmbed tag in the content) wrap be added?

diegoquinteiro:
@kkranzo as of the latest version you can already use `<div class="embed">...</div>` as a generic wrapper. Any iframe should also work too (let me know if it does not).

embeds.php
```diff
@@ -41,8 +41,12 @@ function instant_articles_embed_oembed_html( $html, $url, $attr, $post_ID ) {
 		$provider_name = 'twitter';
 	} elseif( false !== strpos( $providerURL, 'youtube.com' ) ) {
 		$provider_name = 'youtube';
+    } elseif( false !== strpos( $providerURL, 'vimeo.com' ) ) {
```
diegoquinteiro: Please fix indentation to use tabs.

diegoquinteiro:
Indentation looks good, but still need to resolve the conflicts

PR #74, opened 2016-03-29
Adding ability to skip posts

imranglobal:
Added filter to skip posts based on a filter which can be plugged into anything (admin controls or tags etc ).
As per conversation in VIP Ticket#51387

### DIFF
--- a/feed-template.php
+++ b/feed-template.php
@@ -11,6 +11,10 @@ $last_modified = null;
 		<description><?php bloginfo_rss( 'description' ) ?></description>
 		<?php while ( have_posts() ) : the_post(); ?>
 			<?php
+			// Allow to skip posts before loading anything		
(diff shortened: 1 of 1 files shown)


mjangda:
@imranglobal There are a few places where we may want this run, e.g.

- Submitting for review: https://github.com/Automattic/facebook-instant-articles-wp/blob/0bd3899ca4cc5e2bba212ecf30d4d8a733ba0541/wizard/class-instant-articles-wizard-review-submission.php
- Submitting article: https://github.com/Automattic/facebook-instant-articles-wp/blob/9d427f218ac319b76556466f09435a581e5d7311/class-instant-articles-publisher.php#L52
- Displaying metabox

Maybe we should add a new function with the filter (and the post_type and post_status checks) which can be re-used across the board.

diegoquinteiro:
This was implemented (although in the inverse way) by #504.

PR #87, opened 2016-04-06
Clean unused files + fixes for 5.4 compatibility

diegoquinteiro:


YOzaz:
Short array syntax is available from PHP 5.4, so no need to replace it.

diegoquinteiro:
@YOzaz thanks for merging. We're probably try to lower the requirement to 5.3, so I went ahead and replaced the few `[]`s out there.

Can you merge these fixes for 5.4 also?

diegoquinteiro:
@mjangda thank you! fixed =)

PR #90, opened 2016-04-07
Change readme.txt to reflect changes in 2.0

diegoquinteiro:
I've rewritten some readme sections to include:

- Instructions to use composer
- Instructions to use the API integration instead of RSS


readme.txt
```diff
@@ -25,22 +23,21 @@ It's important to note that if you use meta fields to add extra text, images or
...
-
-Facebook will fetch your feed every 3 minutes.
+Facebook requires a minimum number of articles in your feed before they will review it. Once your feed has been approved, new posts will automatically be taken live on Instant Articles, and existing posts will be taken live once you update them.
 
 == Installation ==
 
-1. Upload the folder to the `/wp-content/plugins/` directory
-1. Activate the plugin through the 'Plugins' menu in WordPress
+1. Run `composer install` on the root of the plugin folder. Make sure you have [Composer](https://github.com/composer/composer) installed.
+2. Upload the folder to the `/wp-content/plugins/` directory
```
YOzaz: Shouldn't it be opposite - upload, then run?
Or, "run composer on an environment where it is installed".

TStrothjohann:
Hey, Awesome You have integrated the API! Are you going to include the dependencies in the plugin package later on? Many users don't have shell access to their server and wont't know how to run composer. Thanks!

readme.txt
diegoquinteiro: As a lot of folks don't have SSH access to their web server, it is safer to composer install locally, then upload.

diegoquinteiro:
@TStrothjohann yes, I am going to release a build with the dependencies included today.

PR #95, opened 2016-04-08
Ensure that Related Articles are not stripped

srtfisher:
Allow the `<ul>` element to be empty for Related Articles: https://developers.facebook.com/docs/instant-articles/reference/related-articles

Ensure that `<ul>` elements with a Instant Article class keep their classes and that they are not emptied.

This can probably be done a better way to filter the element and still keep the class attached.


srtfisher:
Closing this out since the latest version of the plugin will support this.

PR #99, opened 2016-04-11
remove audience network placement ID validation

mfelser:
Placement IDs contain an underscore so they are not validating with the
current code

### DIFF
--- a/settings/class-instant-articles-option-ads.php
+++ b/settings/class-instant-articles-option-ads.php
@@ -185,19 +185,7 @@ class Instant_Articles_Option_Ads extends Instant_Articles_Option {
 
 				case 'fan_placement_id':
 					if ( isset( $field_values['ad_source'] ) && 'fan' === $field_values['ad_source'] ) {
-						if (
-							! is_numeric( $field_values[ $field_id ] ) ||
-							strval( absint( $field_values[ $field_id ] ) ) !== $field_values[ $field_id ]
-						) {
-							add_settings_error(
-								$field_id,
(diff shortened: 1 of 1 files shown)


settings/class-instant-articles-option-ads.php
```diff
@@ -185,19 +185,7 @@ public function sanitize_option_fields( $field_values ) {
...
-							add_settings_error(
-								$field_id,
-								'invalid_placement_id',
-								'Invalid Audience Network Placement ID provided'
-							);
-							$field_values[ $field_id ] = $field['default'];
-						} else {
-							$field_values[ $field_id ] = absint( $field_values[ $field_id ] );
-						}
+						$field_values[ $field_id ] = absint( $field_values[ $field_id ] );
```
demoive: WordPress's [`absint()`](https://developer.wordpress.org/reference/functions/absint/) uses PHPs [`intval()`](http://php.net/manual/en/function.intval.php) which would truncate the value of the Placement ID at the underscore.

```php
intval('12345');  # 12345
intval('123_45'); # 123
```

demoive:
@mfelser, thanks for this suggestion. I've placed a comment in your PR.

diegoquinteiro:
Thanks a lot @mfelser! Very good catch.

@mfelser can you incorporate the feedback?

diegoquinteiro:
I've actually fixed this directly on master, @mfelser, I needed to fix it quickly as this was very severe.

demoive:
Confirmed, this PR can be closed.

PR #105, opened 2016-04-12
Update template-settings-info.php

piscis:
https://codex.wordpress.org/Function_Reference/site_url
    The site_url template tag retrieves the site url for the current site (where the WordPress core files reside) 

Incase you're moved the Wordpress core files to its own directory,  Wordpress will suggest that you use `http://[your-domain]/[your-wp-core-files-dir]` instead of `http://[your-domain]/`.

I think to use `home_url()` instead is a better way in this scenario.


diegoquinteiro:
Hi @piscis,

Thanks for your contribution. That makes a lot of sense!

PR #111, opened 2016-04-12
add a category filtering option

mfelser:
Add a GUI to select which categories will be included in the Instant
Articles feed.


diegoquinteiro:
Thank you so much @mfelser, this is great! 

Here you're only filtering the RSS feed. This plugin also publishes articles as you save your posts, so you need also change the `class-instant-articles-publisher.php` and the meta-box to filter out posts in that category.

Also, what you think about putting the setting inside the "Publishing" section, instead of creating a new section?

mfelser:
No problem @diegoquinteiro. I just folded the option into the 'publishing' section and set the publisher to ignore posts not in the approved category list.

Not sure how to handle the meta-box because a user may change the post's categories while editing it.

gabu69:
Could it be added something to blacklist certain categories,

ie We use 1-2 playbuzz quizzes per day but theres still no support for them. Sometimes this articles fall under 2-3 categories but obv one would be "Quizzes", blacklisting quizes wouldnt display them, cause if it displays the other 2 categories it will show still quizzes

ChrisCree:
From a usability standpoint I agree with gabu69 above that it would be good to be able to select categories to omit as well. That way if a new category is added after the plugin is configured the user does not have to remember to go into the Instant Articles setting to add it.

class-instant-articles-publisher.php
```diff
@@ -37,11 +37,25 @@ public static function submit_article( $post_id, $post ) {
 			return;
 		}
 
+
```
everton-rosario: I guess this white line in not needed.

diegoquinteiro:
Hello @mfelser, can you rebase this one to the latest version so we can wrap it up and merge it? This feature is great, thanks for all your work getting through it!

timjacobi:
@diegoquinteiro we can do the merging. Please confirm this is still needed.

PR #115, opened 2016-04-12
Added information about initial publishing

msurguy:
This information is missing AFAIK which creates other trouble (people start using RSS feed to submit)


diegoquinteiro:
Thanks a lot for that. Simple and clear =)

PR #151, opened 2016-04-14
Stop publishing drafts

diegoquinteiro:
We decided to drop the feature where we submit drafts as drafts to Instant Articles. 

The feedback from community is that this behaviour is unexpected and that their libraries got polluted.


everton-rosario:
ship it

diegoquinteiro:
screen shot 2016-04-14 at 20 14 39

PR #161, opened 2016-04-15
[DO NOT MERGE] IA should get unpublished when post type of WP post is changed to unpublished

fabianpimminger:
As a direct consequence of #151, Instant Articles don't get unpublished when the post status of the wordpress post is changed from `published` to something else (`draft`, `future`, `private` ...).

This PR hooks into the `transition_post_status` hook which gets triggered whenever a post status is changed. It only affects posts which previously were published and got unpublished. See https://codex.wordpress.org/Post_Status_Transitions#transition_post_status_Hook for details

### DIFF
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -22,6 +22,7 @@ class Instant_Articles_Publisher {
 	 */
 	public static function init() {
 		add_action( 'save_post', array( 'Instant_Articles_Publisher', 'submit_article' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


diegoquinteiro:
Thanks  a lot for contributing!

timjacobi:
@diegoquinteiro can you check if this is still good to go? I can do the merge manually if @fabianpimminger doesn't respond.

timjacobi:
I checked this and it was written for the old version of the plugin. Leaving this open as reference. Nice to have feature.

PR #173, opened 2016-04-18
Fixed empty link to Instant Articles signup

demoive:
The link was actually empty. This PR adds the link with in a condition for when a Page ID has been defined.


diegoquinteiro:
Nice catch =)

PR #177, opened 2016-04-19
Use H1 for header title and cover image caption

gemedet:
This allows markup inside the header title and cover image caption, e.g.:
```html
<header>
    <figure>
        <figcaption>Some caption <i>in italic</i></figcaption>
    </figure>
    <h1>Some title <b>in bold</b></h1>
</header>
```
Follows facebook/facebook-instant-articles-sdk-php#93


everton-rosario:
Hello @gemedet !
This was a nice catch.

Issue #249, opened 2016-05-10
<div class="embed"> wrapping embeds on non-instant articles. 


kitwestneat:
It appears as if embeds.php strips out all embed filters from other plugins, and adds its own filters, one of which wraps the embed in a &lt;div class="embed"&gt; tag. The problem is that it looks as if embeds.php is run regardless of whether or not the request is for an instant feed.

dashaluna:
What's the status with this?

timjacobi:
Marking this as duplicate of #337 which has more details.

PR #252, opened 2016-05-11
Modify embed filters on the only on Instant Articles feed, not always

kitwestneat:
This PR is to address issue #249

### DIFF
--- a/embeds.php
+++ b/embeds.php
@@ -8,9 +8,12 @@
  */
 
 /**
- * Remove all extra oembed html filters added by themes and plugins.
+ * Initialize the embed filters for the feed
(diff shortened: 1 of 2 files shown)


diegoquinteiro:
This will only handle the RSS integration. What about the API integration (that runs when you save the post)?

timjacobi:
@kitwestneat @diegoquinteiro can we move this forward? The related bug has been reported multiple times.

kitwestneat:
we don't use this plugin anymore, so someone else should lead the charge

GaryJones:
Handled with the solution in #1076.

Issue #198, opened 2016-04-23
Scheduled posts sent empty Instant Articles when published


webadictos:
Hi,

I have an issue with scheduled posts, when they are published i got this error: 

Empty Instant Article: The Instant Article does not have any content.

I need to enter to the wordpress editor and "update" the post so it can be send correctly to Instant Article.

This happen only with scheduled posts.

Thank you,

MLRobbins:
Also seeing this behavior. Any update?

diplatis:
Same here.

MattiaConfalonieri:
Same here... is ridiculous...

Aioros:
So, I can pinpoint the reason for the empty scheduled articles here:

```
protected function _get_the_content() {

global $post, $more;

if ( ! $post ) {
	return '';
}
```
(class-instant-articles-post.php, line 291, v2.10)

The global variable `$post ` is not set up when wp-cron publishes a scheduled post. Right after this block the `$post` variable is checked and properly setup.
That return line seems to be there for some specific reason, but I don't know what it is. Hope this helps.

emanuelet:
Still experiencing the same issue

MattiaConfalonieri:
This is a real pain in the ass guys...

chrisackermann:
Sorry for the headache here. We've confirmed the PR from @Aioros addresses the issue where scheduled posts are not issued to Facebook. We're performing some additional testing to ensure there are no unintended side effects and will plan to merge tomorrow if all's good.

msnadler:
I'm likely late too the party, but I'm having a similar issue with scheduled posts. Though in my case, I don't get the error message others reference, the scheduled posts just don't show up at all.
Unscheduled posts work fine.

chrisackermann:
We've now released Version 2.11 of the plugin which includes the fix provided in #257 from @Aioros (thanks for the contribution!).  Please upgrade to this latest version and let us know if you have any further issues with scheduled posts.

PR #257, opened 2016-05-12
Fix scheduled posts with empty IA content

Aioros:
This PR is to avoid scheduled posts ending up as empty Instant Articles as per issue [#198](https://github.com/Automattic/facebook-instant-articles-wp/issues/198).
The removed check seems to be unnecessary, since a smarter check is done right after more gracefully.

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -290,10 +290,6 @@ class Instant_Articles_Post {
 
 		global $post, $more;
 
-		if ( ! $post ) {
-			return '';
-		}
-
 		// Force $more.
 		$orig_more = $more;
 		$more = 1;


chriszarate:
I'm struggling to understand what purpose that check ever serves. The class is always instantiated with a post object; it should never be relying on global $post. Also, for me, it prevents previewing/checking of draft posts.

Aioros:
Are you referring to the one removed in this PR or the subsequent one? In general, the only use of the global post data that I can think of is in `get_the_kicker()` which calls WP's `get_the_category()` (that should probably be avoided too).

About the drafts, I don't think they are being blocked there, since it's not filtering by `post_status`.

chriszarate:
I meant the one removed by this PR (thank you!). In my use I observe that, in the Post Edit screen, content is only ever retrieved via transients (which are created when the post is published). Otherwise, the check for global $post always returns `''` since the edit screen is not in the Loop.

PR #263, opened 2016-05-13
Escape ampersands before displaying on settings page

gemedet:
This PR allows ampersands in the custom embed code for Ads and Analytics, e.g.:
```
<iframe src="http://foo.com?foo&amp;bar"></iframe>
```

### DIFF
--- a/settings/class-instant-articles-option-ads.php
+++ b/settings/class-instant-articles-option-ads.php
@@ -219,7 +219,7 @@ class Instant_Articles_Option_Ads extends Instant_Articles_Option {
 					if ( isset( $field_values['ad_source'] ) && 'embed' === $field_values['ad_source'] ) {
 						$document = new DOMDocument();
 						$fragment = $document->createDocumentFragment();
-						$field_values[ $field_id ] = str_replace( '&', '&amp;', $field_values[ $field_id ] );
+						$field_values[ $field_id ] = preg_replace( '/&(?!amp;)/', '&amp;', $field_values[ $field_id ] );
 						if ( ! @$fragment->appendXML( $field_values[ $field_id ] ) ) {
 							add_settings_error(
 								'embed_code',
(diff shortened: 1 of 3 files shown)


settings/class-instant-articles-option-ads.php
```diff
@@ -219,6 +219,7 @@ public function sanitize_option_fields( $field_values ) {
 					if ( isset( $field_values['ad_source'] ) && 'embed' === $field_values['ad_source'] ) {
 						$document = new DOMDocument();
 						$fragment = $document->createDocumentFragment();
+						$field_values[ $field_id ] = preg_replace( '/&(?!amp;)/', '&amp;', $field_values[ $field_id ] );
```
everton-rosario: Isn't it better to use esc_url function?

settings/class-instant-articles-option-analytics.php
```diff
@@ -125,6 +125,7 @@ public function sanitize_option_fields( $field_values ) {
 					if ( isset( $field_values['embed_code_enabled'] ) && $field_values['embed_code_enabled'] ) {
 						$document = new DOMDocument();
 						$fragment = $document->createDocumentFragment();
+						$field_values[ $field_id ] = preg_replace( '/&(?!amp;)/', '&amp;', $field_values[ $field_id ] );
```
everton-rosario: Isn't better to use esc_url function?

everton-rosario:
Hey @gemedet, check my inline comments...

Can you check if we have to use esc_url or any other escaping function to make this work?
I guess regular expression replacer isn't a good idea.
https://codex.wordpress.org/Function_Reference/esc_url

Thanks for your PR.

gemedet:
@everton-rosario Thanks for looking it over, I've updated the PR. I think instead of escaping it when saving, we should change how it escapes when displayed in the textarea. If I type `&amp;` into the custom embed code, I'd expect it to display the same when I reload the page.

Currently, `esc_html` leaves `&amp;` untouched as `&amp;`, which gets displayed as `&`.

What we want is `htmlspecialchars` to convert `&amp;` to `&amp;amp;`, which will display as `&amp;`.

everton-rosario:
🚀  Awesome!
LGTM
Thanks! Shipping...

Issue #186, opened 2016-04-20
Ability to control whether a post should be published as an Instant Article


demoive:
There should be the ability to control whether an individual posts/pages should be pushed to Facebook as an Instant Article. Suggestions include:
- Using tags/categories as the condition
- Dedicated UI on each post/page

References:
https://wordpress.org/support/topic/limitexclude-articles-by-category
https://wordpress.org/support/topic/polldaddy-embeds-being-stripped
https://wordpress.org/support/topic/filter-out-custom-post-types-1
https://wordpress.org/support/topic/cancel-article-submission-to-instant-article-on-post-creation
https://wordpress.org/support/topic/how-i-can-filter-out-the-publications-to-be-instant-articles/
https://wordpress.org/support/topic/filter-by-category-14/
https://wordpress.org/support/topic/v30-allow-filtering-of-post-types-used-in-the-feed-how/

https://wordpress.org/support/topic/instant-articles-for-custom-post-type/

kevinlisota:
Please make sure that there is the ability to prevent custom post types from being pushed. We use a bunch of different custom post types that have no reason being shared with FB instant articles.

markbarnes:
It's already possible to do this by modifying the query. I'm doing that to limit by feed to a particular custom post type.

````php
    add_action ('pre_get_posts', array ('modify_ia_query', 11, 1);

    /**
    * Modifies the WordPress query when the instant articles feed is being requested
    * 
    * @param WP_Query $query  The WP_Query object. Passed by reference.
    */
    function modify_ia_query ($query) {
        if (defined('INSTANT_ARTICLES_SLUG') && $query->is_main_query() && $query->is_feed(INSTANT_ARTICLES_SLUG)) {
            $query->set ('post_type', 'em_article');
        }
    }
````

DJUK16:
@markbarnes would that work for example i only want the NEWS category to become an instant articles and other category to just remain normal.

also if so where would i edit to add this code :D

markbarnes:
You should be able to do something similar for categories. Just create a new plugin with that code (although the sample code only works for the feed, you should be able to modify it). I don't have time to help you with the specifics though, I'm sorry.

DJUK16:
way above my head but thank you for the time to replay sir have a wonderful night!

everton-rosario:
This was fixed by PR #504

demoive:
[...]

PR #284, opened 2016-05-20
Disable publishing for custom post types and pages.

chriszarate:
All post types and pages trigger the `save_post` action, but this plugin's post meta box only targets the `post` post type. Publishing other post types and pages goes against expected behavior.

This can be removed if broader support for other post types is implemented as discussed in #186.

### DIFF
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -37,8 +37,19 @@ class Instant_Articles_Publisher {
 			return;
 		}
 
-		// Don't process custom post types or pages.
-		if ( 'post' !== $post->post_type ) {
+		$post_types = array( 'post' );
+
+		/**
+		 * Filter the allowed post types for publishing instant articles.
+		 *
+		 * @since 2.12
+		 *
+		 * @param array $post_types Array of post types to publish.
+		 */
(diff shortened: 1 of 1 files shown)


markbarnes:
This should be filter/action territory IMO. Some of us want to publish Custom Post Types! You can already prevent that by [filtering the query](https://github.com/Automattic/facebook-instant-articles-wp/issues/186#issuecomment-220657524).

chriszarate:
@markbarnes, This doesn't affect feeds. :-) This is in the `submit_article` method of `Instant_Articles_Publisher`, which is used to publish/push articles using the Instant Article API (if you are set up to publish that way).

You're right, though, this should be a filter. This PR was a reaction the undesirable behavior of creating a page or custom post type (or just clicking "update") and having an Instant Article post via the API (perhaps without the user realizing it). It's a bit draconian, so I'll update.

chriszarate:
Looks like this was done in d81dcd4006.

everton-rosario:
@chriszarate @markbarnes Thanks for your inputs.
Yes, both of you were right.

We actually changed the behavior of the plugin, in a way we only submit "post" type to Instant Article.

Feel free to join our group @ Facebook: https://web.facebook.com/groups/629655867187397/

PR #290, opened 2016-05-24
Fix relative URL issue with oembeds not using absolute URLs

stuartshields:
This fixes an issue when oembeds like instagram, twitter etc. use `<script async="" defer="defer" src="//platform.instagram.com/en_US/embeds.js"/>` instead of `<script async="" defer="defer" src="https://platform.instagram.com/en_US/embeds.js"/>`. Without this fix Facebook Instant Articles will report an error to the user about the article.


everton-rosario:
@stuartshields Thanks for this PR.

Sorry for the long wait on reviewing.

Merging it.

PR #295, opened 2016-05-25
added support for multiple ads in embed field

kellemar:
If you're using the "Embed" option for the Ads, separate them in the Embed text box with "::" (without quotes)

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -759,17 +759,26 @@ class Instant_Articles_Post {
 
 			case 'embed':
 				if ( ! empty( $settings_ads['embed_code'] ) ) {
-
-					$document = new DOMDocument();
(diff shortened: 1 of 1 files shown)


class-instant-articles-post.php
```diff
@@ -759,17 +759,26 @@ public function add_ads_from_settings() {
 
 			case 'embed':
 				if ( ! empty( $settings_ads['embed_code'] ) ) {
+					$sectionAds = $settings_ads['embed_code'];
+					$tokenizeAds = explode("::", $sectionAds);
+					//print_r($tokenizeAds);
```
stuartshields: @kellemar You might want to remove this from the PR

kellemar:
Removed the commented print statements.

PR #301, opened 2016-05-27
Check if the global $post object has been set

srtfisher:


class-instant-articles-post.php
```diff
@@ -296,7 +296,7 @@ protected function _get_the_content() {
 
 		// If we’re not it the loop or otherwise properly setup.
 		$reset_postdata = false;
-		if ( $this->_post->ID !== $post->ID ) {
+		if ( empty( $post ) || $this->_post->ID !== $post->ID ) {
```
stuartshields: @srtfisher What instance did you come across to need this?

srtfisher: Adding this plugin to a VIP site caused issues when running unit tests.

PR #302, opened 2016-05-28
Clearer message when "article is being processed"

demoive:
This PR:

* [ ] Abstracts away the two "sub-in-progress" statuses for article submission/updates
* [ ] Adds an new "official" status in the `InstantArticleStatus` class of the SDK (such as `InstantArticleStatus::AWAITING_STATUS_GENERATION`
* [x] Differentiates IA status between "article processing" and "asset ingestion" with updated messages


everton-rosario:
This is not anymore necessary since things are now synchronized.

PR #305, opened 2016-05-31
Allow filtering of the post types used in the feed

### LINKED ISSUE #50: Include other post types
Should we add a filter to our query in our current `pre_get_posts` filter ([here](https://github.com/Automattic/facebook-instant-articles-wp/blob/master/facebook-instant-articles.php#L135)) to add the possibility to filter what post types we put in the feed?

And while we’re at it: What post types should we include by default? Wouldn’t it make sense to include all public post types?

philipjohn:
Fixes #50


everton-rosario:
Ship it.

PR #308, opened 2016-06-06
Fixed error message: Deprecated: Non-static method DOMDocument::loadH…

everton-rosario:
This PR:

* [x] Fixes deprecated call to static DOMDocument::loadHTML() call.

Fixes #

Fixed error message: Deprecated: Non-static method DOMDocument::loadHTML() should not be called statically in /class-instant-articles-post.php on line 617

PR #313, opened 2016-06-07
Fixed statically-called loadHTML()

gemedet:
This PR fixes the deprecated call to static DOMDocument::loadHTML() on line 644.

Follows #308.


everton-rosario:
@gemedet Thnanks for this other PR, it was a warning message we were getting on logs.

Shipping it...

PR #327, opened 2016-06-20
use v2.6 of Facebook API

paulschreiber:
The plugin currently uses v2.5 of the Facebook API. This PR bumps it to v2.6, the current version.


everton-rosario:
@philipjohn @diegoquinteiro double check. Rebase it before shipping.

diegoquinteiro:
looks good, but please rebase before merging

PR #329, opened 2016-06-20
always use HTTPS to load JavaScript SDK; update to 2.6

paulschreiber:
The Facebook SDK should always be loaded over HTTPS. We shouldn't use a protocol-relative URL.

Also bumped the SDK version to 2.6, which is current.


philipjohn:
@diegoquinteiro Do you see any issue with the version bump here?

diegoquinteiro:
@philipjohn no, it is good! :shipit:

PR #341, opened 2016-07-01
Avoid crawler to cache a 404

### LINKED ISSUE #237: Facebook crawler hits while editing post draft and caches a 404
Hi,
since we were still having an old issue where, with this plugin, posts shared on FB often resulted in a cached 404 by the Facebook crawler, I resorted to monitoring the server's access_log to find out when the crawler landed on our posts.
I found out that happens (weirdly) the first time a user opens a draft for editing. As soon as the page is loaded, we receive a visit from "facebookexternalhit" looking for the final URL of the article. The server redirects it to the draft URL (/?p=xxxx) which of course ends up in a 404.

I couldn't track down the reason for such an early visit from the crawler yet. It only (and always) happens when the plugin is active, and it messes with the ability to share the post later, since users have to run it through the FB debugger and scrape it again every time.

Is there a known reason for this behaviour?

diegoquinteiro:
The Facebook Open Graph scraper crawls the URL when trying to get the article ID by URL. In this case, we're avoiding to make this operation for unpublished URLs as it results in crawling a 404.

Fixes #237

### DIFF
--- a/meta-box/class-instant-articles-meta-box.php
+++ b/meta-box/class-instant-articles-meta-box.php
@@ -67,28 +67,30 @@ class Instant_Articles_Meta_Box {
 		Instant_Articles_Settings::menu_items();
 		$settings_page_href = Instant_Articles_Settings::get_href_to_settings_page();
 
-		try {
-			$fb_app_settings = Instant_Articles_Option_FB_App::get_option_decoded();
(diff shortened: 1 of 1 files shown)


diegoquinteiro:
@philipjohn can you please review this PR? =)

PR #343, opened 2016-07-01
Only publish posts (ignores pages and other post types)

diegoquinteiro:


diegoquinteiro:
@philipjohn please take a look at this one too? =)

philipjohn:
We should re-use [the `instant_articles_post_types` filter](https://github.com/Automattic/facebook-instant-articles-wp/blob/4fb6c9cb2c0a18b338d57432b8dbc055a7f92777/facebook-instant-articles.php#L178) here for consistency.

everton-rosario:
:shipit:

PR #348, opened 2016-07-06
Update rules-configuration.json for FB SDK 1.2.0

trepmal:
This PR:

* [x] updates transformer rules

Follows #

Relates to #
https://wordpress.org/support/topic/images-embedded-in-text

Fixes #

The Facebook Instant Articles SDK stable version is now 1.2.0 which updates the sample transformer for WP to accommodate images inside paragraphs
https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/tests/Facebook/InstantArticles/Transformer/CMS/wp-rules.json

### DIFF
--- a/rules-configuration.json
+++ b/rules-configuration.json
@@ -80,6 +80,16 @@
             }
         }
     }, {
+        "class": "ImageInsideParagraphRule",
+        "selector": "//a[img]",
+        "properties": {
+            "image.url": {
+                "type": "string",
+                "selector": "img",
+                "attribute": "src"
+            }
+        }
+    }, {
         "class": "ListItemRule",
         "selector" : "li"
     }, {
--- a/rules-configuration.json
+++ b/rules-configuration.json
@@ -59,9 +59,9 @@
     }, {
         "class": "PassThroughRule",
         "selector": "blockquote p"
-    }, {
+    },{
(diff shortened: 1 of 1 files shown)


everton-rosario:
@trepmal Thanks for this.
We are about to release a new version of the WP plugin which would add the exactly PR.

We rather merge this when we have the new version of WP plugin released. It will happen until end of the week.

trepmal:
Sadly, the recent release is still missing this important rule in the default configuration.

```
{
    "class": "ImageInsideParagraphRule",
    "selector": "img",
    "properties": {
        "image.url": {
            "type": "string",
            "selector": "img",
            "attribute": "src"
        }
    }
}
```
Copying the rule here mainly so passers-by can add this to their custom rules if needed.

Without it there is a high likelihood* the content submitted won't get Facebook's approval. Any text that may be inside a paragraph along with the image will disappear, and Facebook will warn that content doesn't match.

*I don't actually have stats, but it's seems pretty common :)

everton-rosario:
@trepmal Thanks for your report.

I've checked here, and thanks to you we now have it correctly published.

Fix is in this PR: #360

everton-rosario:
@trepmal your contributions are fundamental to Facebook Instant Articles!

PR #353, opened 2016-07-11
Store and use submission ID when available to fetch the status

diegoquinteiro:
This PR:

- Stores the submission ID of the posted article on post metadata.
- Uses the stored submission ID when available to fetch the article status instead of always getting it from the URL.

This fixes an issue where the meta-box does not show that the article was submitted unless you refresh the edit post screen, which feels broken.

### DIFF
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -18,6 +18,11 @@ use Facebook\Facebook;
 class Instant_Articles_Publisher {
 
 	/**
+	 * Key to store the submission status ID on meta data
+	 */
+	 const SUBMISSION_ID_KEY = 'instant_articles_submission_id';
+
+	/**
 	 * Inits publisher.
 	 */
 	public static function init() {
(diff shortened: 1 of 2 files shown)


diegoquinteiro:
@philipjohn if you can, it would be awesome to include this one on 3.0 tomorrow (but let's not delay the release because of it if you can't).

class-instant-articles-publisher.php
```diff
@@ -93,10 +98,12 @@ public static function submit_article( $post_id, $post ) {
...
 				try {
 					// Import the article.
-					$client->importArticle( $article, $take_live );
+					$submission_id = $client->importArticle( $article, $take_live );
+					update_post_meta( $post_id, self::SUBMISSION_ID_KEY, $submission_id );
 				} catch ( Exception $e ) {
 					// Try without taking live for pages not yet reviewed.
-					$client->importArticle( $article, false );
+					$submission_id = $client->importArticle( $article, false );
+					update_post_meta( $post_id, self::SUBMISSION_ID_KEY, $submission_id );
```
everton-rosario: Does this try/catch works well?
Is it needed to try twice?
I know it was here before,  but looks weird.
Its ok to just file an issue so we can double check it later.

everton-rosario:
Take a look on my inline comment.

LGTM

class-instant-articles-publisher.php
diegoquinteiro: When we try to publish, we send it with the flag to take it live. If the page is still pending review, this particular call will fail and an Exception is thrown. We then try to send it without the flag.

When the page is under review, we always send the request twice. But when the page is approved, we always send a single request.

We could, instead, check the review status before and include the flag or not based on the status. But that would mean 2 requests are made (checking the status then sending) in both scenarios. It is a clearer solution, but the current solution performs better on the average case.

We can discuss it later, it is not on the scope of this PR.

everton-rosario:
:ship

PR #356, opened 2016-07-12
Ensure that relative URL checking has a host and a path.

srtfisher:
This PR solves an issue with PHP throwing an `Notice:  Undefined index: host in /var/www/lin/wp-content/themes/vip/plugins/facebook-instant-articles-3.0/class-instant-articles-post.php on line 691`.


everton-rosario:
@srtfisher Thanks for this!
It will get in on next version 3.1

PR #358, opened 2016-07-13
Revert "escaping and sanitization fixes"

diegoquinteiro:
This reverts commit 232b2c9c734856c4e590eb329fd139a2fd5e8b3f.

The escaping here is breaking the UI, as the variables being escaped here DO contain HTML, although it is not user-generated HTML, but hardcoded HTML configured on the descriptions provided for each field.

instant_articles_settings_ _diego_quinteiro_ _wordpress

### DIFF
--- a/settings/class-instant-articles-option-ads.php
+++ b/settings/class-instant-articles-option-ads.php
@@ -134,7 +134,7 @@ class Instant_Articles_Option_Ads extends Instant_Articles_Option {
 
 		?>
 		</select>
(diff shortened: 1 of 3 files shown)


diegoquinteiro:
@philipjohn can we get this out? =)

philipjohn:
I've restored escaping, but using `wp_kses_post()` which will allow the limited HTML we want, and I've also restored the changes in wpcom-helper.php that are still valid.

PR #360, opened 2016-07-13
Improve rules for Images inside Paragraphs and Interactive elements

everton-rosario:
This PR:

* [x] Fixes issue where text inline with images are being dropped.
* [x] Properly fetches width and height for Interactive elements


diegoquinteiro:
:shipit:

Issue #361, opened 2016-07-13
Figcaption for Interactive / Embeds are getting stripped out


krispy1298:
From the [Docs](https://developers.facebook.com/docs/instant-articles/reference/embeds), check out the last example.

In my build, the source markup is...
```html
<div class="pa-interactive">
<figure class="op-interactive"><iframe width="1600" height="900" width="200" height="150" src="https://www.youtube.com/embed/FWO5Ai_a80M?feature=oembed" frameborder="0" allowfullscreen></iframe>
<figcaption>asdfsdfsdfs — dfasdfsdfasdf asfasdf asd fasdfasdfsdfasdfsdfas</figcaption>
</figure>
</div>
```
...and the Transformed Markup is:
```html
<figure class="op-interactive">
  <iframe src="https://www.youtube.com/embed/FWO5Ai_a80M?feature=oembed" class="no-margin" width="1600" height="900"></iframe>
</figure>
```

My custom tranformer Json includes:
```json
        {
        "class": "InteractiveRule",
        "selector" : "div.pa-interactive",
        "properties" : {
            "interactive.url" : {
                "type" : "string",
                "selector" : "iframe",
                "attribute": "src"
            },
            "interactive.width" : {
                "type" : "int",
                "selector" : "iframe",
                "attribute": "width"
            },
            "interactive.height" : {
                "type" : "int",
                "selector" : "iframe",
                "attribute": "height"
            },
            "interactive.iframe" : {
                "type" : "children",
                "selector" : "iframe"
            },
            "interactive.caption" : {
                "type" : "element",
                "selector" : "figcaption"
            }
        }
```

# Expected Result

```html
<figure class="op-interactive">
  <iframe src="https://www.youtube.com/embed/FWO5Ai_a80M?feature=oembed" class="no-margin" width="1600" height="900"></iframe>
  <figcaption>
    This is the caption
  </figcaption>
</figure>
```

# Actual Result

```html
<figure class="op-interactive">
  <iframe src="https://www.youtube.com/embed/FWO5Ai_a80M?feature=oembed" class="no-margin" width="1600" height="900"></iframe>
</figure>
```

---

** Version Info **

* Plugin version: 3.0.1 
* WordPress version: 4.5.3
* PHP version: 5.6.2

everton-rosario:
Hi @krispy1298.

You will need to remove this:

```json
            "interactive.caption" : {
                "type" : "element",
                "selector" : "figcaption"
            }
```

[...]

PR #362, opened 2016-07-13
Removed interactive.caption property from rule InteractiveRule

everton-rosario:
This PR:

* [x]  Cleans up the configuration property interactive.caption from rules, since it is not being used anymore

Relates to #361


everton-rosario:
This is not needed anymore.

PR #367, opened 2016-07-15
Add subtitle to $header when rendering post to IA

jacobarriola:
The plugin has functions built-in for subtitles; however it does not include it when constructing the header. Including it here allows users to use the built-in `instant_articles_subtitle` filter that allows users to target their subtitles, should they be part of their theme.

We’ve run into an issue during the submission process where subtitles where missing and Facebook mandating that we include them.


everton-rosario:
@jacobarriola nice catch!

PR #371, opened 2016-07-18
Migrated take_live parameter to the published

everton-rosario:
This PR:

* [x] Changes take_live to use published parameter

Relates to Deprecation of field **take_live**

It was already renamed on the SDK, but was still lingering on the WordPress plugin.


diegoquinteiro:
:shipit:

PR #375, opened 2016-07-21
Commit blocker option - No warnings in transformer to publish

everton-rosario:
This PR:

* [x] Creates a flag in settings "block_publish_with_warnings"
* [x] Flag defaults to false (disabled)
* [x] If flag is true and transformer results in warnings, Instant Article is not registered 

Depends on SDK: https://github.com/facebook/facebook-instant-articles-sdk-php/pull/154

### DIFF
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -89,6 +89,17 @@ class Instant_Articles_Publisher {
 					$dev_mode
 				);
 
+				// Don't process if contains warnings and blocker flag for transformation warnings is turned on.
+				if ( count( $adapter->transformer->getWarnings() ) > 0
+				  && isset( $publishing_settings['block_publish_with_warnings'] )
+					&& $publishing_settings['block_publish_with_warnings'] ) {
+
+					// Unpublishes if already published
+					$client->removeArticle( $article->get_canonical_url() );
+					delete_post_meta( $post_id, self::SUBMISSION_ID_KEY );
+					return;
+				}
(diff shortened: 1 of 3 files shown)


rules-configuration.json
```diff
@@ -25,7 +25,7 @@
         "selector" : "span"
     }, {
         "class": "ParagraphRule",
-        "selector": "p"
+        "selector": "pad"
```
diegoquinteiro: that doesn't look right ;)

class-instant-articles-publisher.php
```diff
@@ -89,6 +89,17 @@ public static function submit_article( $post_id, $post ) {
 					$dev_mode
 				);
 
+				// Don't process if contains warnings and blocker flag for transformation warnings is turned on.
+				if ( count( $adapter->transformer->getWarnings() ) > 0
+				  && isset( $publishing_settings['block_publish_with_warnings'] )
+					&& $publishing_settings['block_publish_with_warnings'] ) {
+
+					// Unpublishes if already published
+					$client->removeArticle( $article->get_canonical_url() );
```
diegoquinteiro: love this; let's prevent outdated posts to be live =)

diegoquinteiro:
overall it is good, but please fix you test leftover =)

rules-configuration.json
everton-rosario: #fuuuu 
that was a test case. 
Tks for spotting it :-)

PR #379, opened 2016-07-25
Don't call wpautop directly when trying to prepare content for transformation.

rinatkhaziev:
Instead check whether wpautop is attached for 'the_content' and if it's not, add it. The reason for this is that we applied the_content filters already, and the shortcodes were rendered. Calling wpautop after shortcodes are rendered will mangle most of them, unless shortcode returns 1 line of code.

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -311,6 +311,11 @@ class Instant_Articles_Post {
 		 * @since 0.1
 		 * @param string  $content  The current post content.
 		 */
+
+		// Some people choose to disable wpautop. Due to the Instant Articles spec, we really want it in!
+		if ( ! has_filter( 'the_content', 'wpautop' ) )
+			add_filter( 'the_content', 'wpautop' );
+
 		$content = apply_filters( 'the_content', $content );
 
 		// Maybe cleanup some globals after us?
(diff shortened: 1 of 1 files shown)


everton-rosario:
@rinatkhaziev how are the reproducible steps for the problem you've pointed out?

rinatkhaziev:
Hey @everton-rosario, sure thing, here's quick and dirty test 


```php
add_action( 'init', function() {
	$test_content = "
	About wpautop
	About wpautop2

	About wpautop3

	About wpautop4

	[test-sc]
		";

	$filtered = apply_filters( 'the_content', $test_content ) ;
	var_dump( "input string: \n\n" . $test_content );
	var_dump( "Expected output: \n\n" . $filtered, "\n\n", "wpautop'd after shortcode \n\n" . wpautop( $filtered ) );
	exit;

});

add_shortcode( 'test-sc', function( $atts, $content ) {
	return "
	<script>

	console.log('autop');

	console.log('autop');

	console.log('autop');

	</script>
	";
} );
```

and the output: 
```
string 'input string: 


About wpautop
About wpautop2

About wpautop3

About wpautop4

[test-sc]
	' (length=90)
string 'Expected output: 

<p>About wpautop<br />
About wpautop2</p>
<p>About wpautop3</p>
<p>About wpautop4</p>

<script>

console.log('autop');

console.log('autop');

console.log('autop');

</script>
	
' (length=197)
string '

' (length=2)
string 'wpautop'd after shortcode 

<p>About wpautop<br />
About wpautop2</p>
<p>About wpautop3</p>
<p>About wpautop4</p>
<p><script></p>
<p>console.log('autop');</p>
<p>console.log('autop');</p>
<p>console.log('autop');</p>
<p></script></p>
' (length=234)
```

PR #385, opened 2016-07-29
Implemented compat for the get-the-image plugin

### LINKED ISSUE #381: Image in post not displaying

1. 
2. 
3. 

# Expected Result

*Pulling image from the article and publish in instant article.

# Actual Result

* I'm getting an instant article without image even though i input image in the article.
Though, image will pick up only when input in feature attachment

---

** Version Info **

* Plugin version: 3.0.1
* WordPress version:  wordpress 4.5.3
* PHP version:  5.5

everton-rosario:
This PR:

* [x] Adds configuration rules for the get-the-image plugin
* [x] Compat layer setup with the new plugin

Fixes #381

### DIFF
--- a/compat.php
+++ b/compat.php
@@ -7,6 +7,11 @@
  * @package default
  */
 
+if ( !function_exists( 'is_plugin_active' ) ) {
+	require_once ( ABSPATH . '/wp-admin/includes/plugin.php' );
+}
+
+
 // Load compat layer for Co-Authors Plus.
 if ( function_exists( 'get_coauthors' ) && ! defined( 'CAP_IA_COMPAT' ) ) {
 	include( dirname( __FILE__ ) . '/compat/class-instant-articles-co-authors-plus.php' );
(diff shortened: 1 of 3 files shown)


compat.php
```diff
@@ -7,6 +7,11 @@
  * @package default
  */
 
+if ( !function_exists( 'is_plugin_active' ) ) {
```
diegoquinteiro: Why do we need this now?

compat/get-the-image-rules-configuration.json
```diff
@@ -0,0 +1,13 @@
+{
+  "rules": [{
+      "class": "ImageInsideParagraphRule",
+      "selector": "figure.wp-caption",
```
diegoquinteiro: Did you test to confirm the caption is being read too?

compat.php
everton-rosario: This should be how we check if any plugin is active.

compat/get-the-image-rules-configuration.json
everton-rosario: Yes.

adavize4u:
Still having little problem:
1. Sometimes, text striped from the transform post, making our facebook update useless.
2. Image mostly appear when article published with new image, but old image picked from the site uploads, it sometimes not displaying image.

everton-rosario:
@adavize4u Can you share the markup sent to transformer, for the cases you've mentioned?

To get the transformation code, just click in the link "**[Toggle debug information]**" in the "Instant Articles" meta-box and then copy and past the both editor's text: "**Source Markup**" and "**Transformed Markup**"

This way I can reproduce and check where are the corner cases.

Thanks, @adavize4u

everton-rosario:
Hey @philipjohn/@diegoquinteiro, why this PR didn't get thru to the release 3.1?

philipjohn:
Playing Milestone tennis :)

PR #383, opened 2016-07-27
Fix updating published posts result in error

chrisadas:
#This PR:

* [x] Fixed error that occurs when updating a post that is already published, which prevent the changes from being published (it still gets saved as revision)

Fixes this issue #382 

Issue was intermittent and could be related to whether Instant Articles are under review on Facebook side (and thus disabling ability to delete Instant Articles on Facebook)

Fix suggested by @everton-rosario and worked for me.

PR #387, opened 2016-08-01
Fix loadHTML warning

gemedet:
This PR hides warnings when using loadHTML, with a fix suggested by @everton-rosario.

Follows #383.


everton-rosario:
Thanks @gemedet

PR #389, opened 2016-08-02
Redirect settings

everton-rosario:
This PR:

* [x] Redirect to settings page when plugin gets active.


diegoquinteiro:
:shipit:

PR #392, opened 2016-08-04
Load functions which rely on `plugins_loaded` action hook in wp.com e…

nprasath002:
Load functions which rely on `plugins_loaded` action hook in wp.com environment where it has been already fired


rinatkhaziev:
This should probably go into wpcom-helper.php, but I approached it with 
`add_action( 'instant_articles_before_transform_post', 'instant_articles_load_compat' );` so far seems to be working fine.

PR #394, opened 2016-08-05
Apply "the_title" filter

### LINKED ISSUE #372: qTranslate X with Facebook Instant Article
Hi,

We are using qTranslate X for a multi-language wordpress site.

When I post the article and to using "Instant Articles for WP" auto publish to Facebook and I found the all language tags appear in the title.

screenshot_20160720-144630

But the content only show single language seems normally.

How can I fix the title with only single language?

Thanks!

yssk22:
This PR:

* [x] apply 'the_title' filter to get the filtered title string on Instant Article (same as the content where we apply 'the_content' filter.

Fixes #372

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -85,6 +85,14 @@ class Instant_Articles_Post {
 		$title = $this->_post->post_title;
 
 		/**
+		 * Apply the default filter 'the_title' for the post tite.
+		 *
+		 * @since 3.1
+		 * @param string  $title  The current post title.
+		 */
+		$title = apply_filters( 'the_title', $title);
+
+		/**
 		 * Filter the post title for use in instant articles.
 		 *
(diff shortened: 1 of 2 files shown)


everton-rosario:
@yssk22 Awesome work!

Merging...

tennyy:
Thanks for your help!

After I patched the file with "class-instant-articles-post.php", the title kept in one language, but the   category will become in both language.

poke

I using quick edit to update all post to Facebook, but when I using normal edit post and save again, everything seems fine.

Can you help me to solve the title problem?

Many Thanks!

yssk22:
@tennyy you need to update all posts you already have in Instant Article since my fix doesn't change anything on existing articles. 

As for the category, I think it should not happen in the recent version of the plugin (since the code use the filtered string of categories).

Issue #344, opened 2016-07-01
UTF8 not working or different language not working


dhakalsuresh:
1.Test in different language is not working, i just see text as "?????????????????"  format, which is completely not readable
- 
  ![capture](https://cloud.githubusercontent.com/assets/8805646/16529182/940c3332-3feb-11e6-8f21-ed344d64386d.PNG)

*\* Version Info **
- Plugin version: 2.11 
- WordPress version: 4.5.3
- PHP version: 5.5.35

philipjohn:
Could you provide some more information on the steps to reproduce this please?

Are you using any translation plugins?

Are you using a specific locale for your WordPress install?

Also, please provide the exact versions - "latest" will change, so we need the exact versions this occurs in at the time of reporting.

dhakalsuresh:
I'm just using some front-end plugins,
SoundCloud Shortcode,
WP Subtitle,
Image Upload Widget,
Responsive Menu

I tried to remove all plugins but it's looks same . here is the link how it's looks in rss feed: http://nepalsandesh.com/feed/instant-articles

Wordpress Version : Version 4.5.3

sakatam:
Hi @dhakalsuresh, I have been trying to reproduce this but with no luck. Could you provide detailed info about the database as well? Please run the following commands on mysql cli, and put the result here? 

```sql
SHOW VARIABLES LIKE "version%";
```

```sql
SHOW CREATE TABLE wp_posts;
```

Thank you in advance for your help!

dhakalsuresh:
if i replace `render ` function in `feed-template.php` file with `get_the_content()` , it's fetch the real data.
`<![CDATA[<?php echo $instant_article_post->to_instant_article()->render(); ?>]]>`

Hello @sakatam, Here is what i saw with your query.
1. 
`version	5.5.43-37.2-log
version_comment	Percona Server (GPL), Release rel30.2, Revision 38...
version_compile_machine	x86_64
version_compile_os	Linux`

2. `CREATE TABLE `wp_posts` (
 `ID` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
 `post_author` bigint(20) unsigned NOT NULL DEFAULT '0',
 `post_date` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
 `post_date_gmt` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
 `post_content` longtext NOT NULL,
 `post_title` text NOT NULL,
[...]

PR #395, opened 2016-08-08
introduce portable-utf8 lib

### LINKED ISSUE #384: Arabic articles appear as a question marks (??????)
1. 
2. 
3. 
# Expected Result
- Showing Arabic letters normally
# Actual Result
- Articles written in Arabic, appears in this form : ????? ????? ???????? ???? ?????? ???? ????

---

*\* Version Info **
- Plugin version: 3.0.1
- WordPress version: 4.5.3
- PHP version: PHP 5.4.45

sakatam:
This PR:

- [x] adds [`portable-utf8 library`](https://github.com/voku/portable-utf8) so multi-byte chars get converted properly even when mbstring is not available.

Fixes #384 #344

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -17,6 +17,8 @@ use Facebook\InstantArticles\Elements\Image;
 use Facebook\InstantArticles\Elements\Caption;
 use Facebook\InstantArticles\Elements\Footer;
 use Facebook\InstantArticles\Transformer\Transformer;
+use voku\helper\UTF8;
+
 /**
  * Class responsible for constructing our content and preparing it for rendering
  *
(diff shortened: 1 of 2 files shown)


sakatam:
I am not quite sure if it is ok to bring extra dependencies. but I believe this is the only way to fix the garbled text issue without using `mbstring` on plugin side.

the root cause of this issue is the fact that php's DOMDocument does not handle multi-byte chars properly. the ideal solution that I think is to encapsulate the mb string conversion in the SDK so consumer do not need to take care about it.

philipjohn:
@diegoquinteiro Should we shift this to the SDK assuming the last comment is correct on the ideal solution>

sakatam:
closing this PR as we have started working on the ideal solution.

PR #399, opened 2016-08-10
Revert "Redirect settings"

philipjohn:
Reverts Automattic/facebook-instant-articles-wp#389

While it's likely that a user will want to configure the plugin after activation, we shouldn't assume, and certainly shouldn't interrupt WP's normal behaviour.

A better approach would be a one-time post-activation admin message to point people to the settings page.


everton-rosario:
@philipjohn Do you have an example about this correct approach on "one-time post activation"?

philipjohn:
@everton-rosario Sure, the best way would be to hook into the `admin_notices` action whilst in the `register_activation_hook` action. [This blog post](https://tommcfarlin.com/plugin-activation-message/) has a good example.

PR #401, opened 2016-08-10
Only load compat layers when we're rendering a feed

philipjohn:
The plugin currently does not load compat layers conditionally.

One result is that embeds (e.g. for Facebook) are transformed into FBIA markup for normal front-end posts. This means embeds don't render because they are not recognisable HTML.

This PR moves the loading of compat layers to the feed initiation so that we only load the compatibility code when we are definitely rendering the feed.

### DIFF
--- a/compat/class-instant-articles-co-authors-plus.php
+++ b/compat/class-instant-articles-co-authors-plus.php
@@ -18,7 +18,10 @@ class Instant_Articles_Co_Authors_Plus {
 	 * Init the compat layer.
 	 */
 	function init() {
-		add_filter( 'instant_articles_authors', array( $this, 'authors' ), 10, 2 );
+		// Only make changes when we're building an FBIA response
+		if ( instant_articles_is_fbia_response() ) {
+			add_filter( 'instant_articles_authors', array( $this, 'authors' ), 10, 2 );
+		}
 	}
 
 	/**
--- a/compat/class-instant-articles-jetpack.php
+++ b/compat/class-instant-articles-jetpack.php
@@ -13,9 +13,12 @@ class Instant_Articles_Jetpack {
 	 *
 	 */
 	function init() {
-		$this->_fix_youtube_embed();
-		$this->_fix_facebook_embed();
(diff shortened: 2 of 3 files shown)


philipjohn:
@diegoquinteiro @everton-rosario Would you sanity check this for me please? I'm unsure if this will negatively affect the compatibility layers being applied when FBIA markup is generated other than for the feed.

diegoquinteiro:
Hey @philipjohn; Yes, unfortunately that would fail to include the compat layers on the API submission. Some compat files also add options to the settings screen (for instance, the Google Analytics compat file). Can we require the files also on the admin screens?

philipjohn:
Damn, there was my hoping for a simple fix :)

We ought to help ourselves out with some sort of helper function. E.g. `is_fbia_response()` - or similar - would return true when we *should* be returning FBIA-compatible output.

Any other time, such as normal front-end loads, it would return false. That way, we don't alter markup outside of responses that need FBIA markup.

I feel like this is a pretty serious bug that we need to get into 3.1 :) I'll commit a helper function, and invoke it for feeds as an example but I'm unsure of the best place to do this for other responses - could you help with that please @diegoquinteiro ?

philipjohn:
Okay, those commits have a very rough implementation of what I'm thinking. Completely untested as yet - let me know what you think!

diegoquinteiro:
Hey @philipjohn, actually we don't need to put these checks on every compat. If you look closely, they only hook to the own plugin filters, not changing anything on the output for regular page renderings. The only actual problem happens in the embed.php file, where we are mangling the output in every situation. I'll put another PR with my ideas to fix that.

diegoquinteiro:
That looks like a good approach, thanks! But, still, this is a harder problem. For instance, there is the embeds.php file, which is not guarded by this method you created. I will take a closer look at this issue, and I suspect I'll have to refactor some core methods.

I'm pushing this to 3.1.1 so we can release 3.1 at least on GitHub today. I'll need to take some more time this week to test and finalize this solution for next week.

philipjohn:
Closing in favour of #457

PR #413, opened 2016-08-17
Fix lingering reference to the old settings page on meta-box

diegoquinteiro:
This PR removes a lingering reference to the old settings page removed on 3.1. This was breaking the meta-box.


diegoquinteiro:
@philipjohn also merging this one because that's an obvious error we need to fix. Please take a look too.

PR #414, opened 2016-08-18
WordPress.com VIP Improvement for Text Domain and Compatibility Modules

rleeson:
This PR:

* Adds environment detection as a constant (IA_PLUGIN_IS_WPCOM_ENV) for WP.COM VIP sites 
* Hooks text domain and third party module compatibility load to after_setup_theme for VIP sites based on the added constant

Relates to #392 

Removes the possibility of the hooks being called multiple times and works with other filters, like those defined in #401.


philipjohn:
Thanks Ryan!

This workaround is probably better off living in `wpcom-helper.php` rather than in the plugin itself, so that we don't have wpcom-specific code in the wporg plugin. However, as mentioned in #401 it looks like we probably need something deeper to account for all the different compats we have in.

PR #417, opened 2016-08-23
Support for Playbuzz plugin

### LINKED ISSUE #134: Compatibility with playbuzz plugin and embed?
Is there any future support for playbuzz plugin and embeds?

https://wordpress.org/plugins/playbuzz/
https://publishers.playbuzz.com/how_to/embed-facebook-instant-article/

At the moment it wont display anything and gotta unpublish them manually all the time

everton-rosario:
This PR:

* [x] Adds compat support for the playbuzz plugin
* [x] Adds setup rules for the playbuzz markup
* [x] Checks to enable these rules only if playbuzz is enabled.

Fixes #134

### DIFF
--- a/compat.php
+++ b/compat.php
@@ -44,6 +44,7 @@ if ( is_plugin_active( 'get-the-image/get-the-image.php' ) ) {
 	include( dirname( __FILE__ ) . '/compat/class-instant-articles-get-the-image.php' );
 	$gti = new Instant_Articles_Get_The_Image;
 	$gti->init();
+}
 
 // Load support for Playbuzz plugin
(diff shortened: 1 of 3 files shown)


compat.php
```diff
@@ -46,3 +45,10 @@
 	$gti = new Instant_Articles_Get_The_Image;
 	$gti->init();
 }
+
+// Load support for Playbuzz plugin
+if ( is_plugin_active( 'playbuzz/index.php' ) ) {
```
philipjohn: This isn't going to work on VIP (which is probably more our problem, but we should fix it here anyway) but also is vulnerable to changes in the playbuzz plugin itself.

It'd be more reliable to check for the existence of a class or function, if possible.

PR #418, opened 2016-08-24
Call instant_articles_post_published hook to determine published status

apeschar:
This adds a `instant_articles_post_published` hook to allow a plugin to change the `published` field on newly saved posts.

### DIFF
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -122,7 +122,7 @@ class Instant_Articles_Publisher {
 					$published = false;
 				} else {
 					// Any publish status other than 'publish' means draft for the Instant Article.
-					$published = apply_filters( 'instant_articles_post_published', true, $post->ID );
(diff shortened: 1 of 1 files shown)


class-instant-articles-publisher.php
```diff
@@ -104,7 +104,7 @@ public static function submit_article( $post_id, $post ) {
 					$published = false;
 				} else {
 					// Any publish status other than 'publish' means draft for the Instant Article.
-					$published = true;
+					$published = apply_filters( 'instant_articles_post_published', true, $post->ID );
```
diegoquinteiro: You need to use the `$post_id` variable in this context:
```php
$published = apply_filters( 'instant_articles_post_published', true, $post_id );
```

diegoquinteiro:
Thanks for your contribution, it looks good! You just need to update the code and rename the variable before we merge it (look at my review).

apeschar:
Done.

PR #433, opened 2016-09-01
fix can't use function return value in write context

donaldG:
This PR:
- [x] 
- [ ] 
- [ ] 

Follows #

Relates to #

Fixes #

Fixes:

Fatal error: Can't use function return value in write context in
/var/www/site.com/wp-content/plugins/fb-instant-articles/wizard/template
s/wizard-template.php on line 101

Causes the edit screen not to show, empty() prefers variable not
function

### DIFF
--- a/wizard/templates/wizard-template.php
+++ b/wizard/templates/wizard-template.php
@@ -98,7 +98,9 @@
 
 <?php if ( ! $ajax ) : ?>
 	</div>
-	<?php if ( ! empty( get_settings_errors() ) ) : ?>
+	<?php 
+	$get_settings_errors = get_settings_errors();
+	if ( ! empty( $get_settings_errors ) ) : ?>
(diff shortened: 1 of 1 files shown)


donaldG:
This *should* address issue #438. I'm not seeing this issue locally but on our dev/stage/production environments I'm seeing this issue and it appears I'm not the only one. For anyone needing to edit the settings this is a fix that needs to happen pretty quick. I'm currently trying to get our articles approved (I've got one minor fix left) but I need to edit the custom transformer rules and I can't get to them b/c of this error.

LeBaux:
Can confirm the fix on the WP 4.6.1

donaldG:
Beautiful! Thanks!

diegoquinteiro:
Hey @donaldG, I think we "accidentally" fixed it in a previous version by replacing empty() with count(). Thanks a lot for your contribution =)

PR #435, opened 2016-09-02
Setup Wizard copy updates

demoive:
This PR:

* [x] Updates copy within Steps 1 & 3 of the Setup Wizard
* [x] Removes the "Enabled" label next to available FB Pages for selection
* [x] Uses CSS to store the state of the "advanced settings toggle" rather than manipulating DOM content in JS
  * [x] The also permits the use of ▶︎ for the "closed" state to match the style of the corresponding "open" state (▼). Previously, ► was used to avoid it from being replaced with the :arrow_forward: emoticon by the built-in WordPress styling.
  * [ ] The `slideToggle()` animation for opening/closing the section no longer exists


css/instant-articles-wizard.css
```diff
@@ -350,18 +350,27 @@ a.instant-articles-button {
 	font-weight: bold;
 }
 .instant-articles-advanced-settings {
```
demoive: Selector now targets a <p> instead of an <a> tag so `text-decoration` and `display` properties aren't necessary.

js/instant-articles-wizard.js
```diff
@@ -122,15 +122,14 @@ function instant_articles_wizard_bind_events () {
 jQuery( document ).ready( function () {
 	instant_articles_wizard_bind_events();
 
-	jQuery( '.instant-articles-wizard-toggle' ).on( 'click', function () {
-		var text = jQuery( this ).text();
-		if ( text.indexOf( '►' ) !==  -1 ) {
-			text = text.replace( '►', '▼' );
+	jQuery( '.instant-articles-wizard-toggle a' ).on( 'click', function () {
+		$advancedSettingsContainer = jQuery('.instant-articles-advanced-settings');
```
everton-rosario: Spacing is broken here. 

Should be:
`jQuery( '.instant-articles-advanced-settings' );`

js/instant-articles-wizard.js
```diff
@@ -122,15 +122,14 @@ function instant_articles_wizard_bind_events () {
...
+		if ( $advancedSettingsContainer.attr('data-state') === 'closed' ) {
+			$advancedSettingsContainer.attr('data-state', 'opened');
```
everton-rosario: Review the spacing to be like this:

$var.function( 'parameter spaced', 'another param' );

css/instant-articles-wizard.css
everton-rosario: Just check my inline comments.

everton-rosario:
@demoive Take a look on the conflicts. Probably you need to rebase it and then fix conflicts, before making it ready.

wizard/templates/page-selection-template.php
```diff
@@ -43,12 +43,10 @@
...
 						<label>
 							<?php echo esc_html( $page[ 'page_name' ] ) ?>
-							<?php if ( $page[ 'supports_instant_articles' ] ) : ?>
-								<span class="page-enabled">✔ Enabled</span>
-							<?php else : ?>
+							<?php if ( ! $page[ 'supports_instant_articles' ] ) : ?>
 								<span class="page-not-enabled">
-									This page has not been signed up yet.
-									<a href="https://www.facebook.com/instant_articles/signup?redirect_uri=<?php echo urlencode( $settings_url ) ?>&page_id=<?php echo urlencode( $page[ 'page_id' ] ) ?>">Sign Up</a>.
+									First, <a href="https://www.facebook.com/instant_articles/signup?redirect_uri=<?php echo urlencode( $settings_url ) ?>&page_id=<?php echo urlencode( $page[ 'page_id' ] ) ?>">sign up</a>
```
philipjohn: We should use `rawurlencode()` over `urlencode()` here

demoive: I've looked at the difference between those two functions, and to be honest, I don't know the specs (e.g. RFC 3986) well enough to be able to determine which circumstances merits each one. Any insight there?

I am updating the PR with your request.

PR #436, opened 2016-09-02
Wp tests migration

everton-rosario:
This PR:

* [x] Configures phpunit
* [x] Moves phpunit tests about WP that was on the SDK that should be here.

Relates to #


diegoquinteiro:
lgtm, @philipjohn can you double check?

PR #437, opened 2016-09-02
Blocks publishing of articles with transformation warnings

everton-rosario:
This PR:

* [x] Enables the blocker if article contains Warnings.

### DIFF
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -101,7 +101,7 @@ class Instant_Articles_Publisher {
 
 				// Don't process if contains warnings and blocker flag for transformation warnings is turned on.
 				if ( count( $adapter->transformer->getWarnings() ) > 0
-				  && ( ( ! isset( $publishing_settings[ 'publish_with_warnings' ] ) ) || ! $publishing_settings[ 'publish_with_warnings' ] )
+				  && ( ! isset( $publishing_settings[ 'publish_with_warnings' ] ) || ! $publishing_settings[ 'publish_with_warnings' ] )
 					&& ( ! $force_submit )
 					) {
 
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -21,6 +21,7 @@ class Instant_Articles_Publisher {
 	 * Key to store the submission status ID on meta data
 	 */
 	 const SUBMISSION_ID_KEY = 'instant_articles_submission_id';
+	 const FORCE_SUBMIT_KEY = 'instant_articles_force_submit';
 
 	/**
 	 * Inits publisher.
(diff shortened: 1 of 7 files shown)


diegoquinteiro:
This will enable this option for everyone who have not set the option before. I'm not against it, but we should:

- Put a message on the meta-box explaining the behavior.

- Proper handle this flag when submitting for review. Right now we send the last 5 articles for review regardless of their statuses. When changing this flag to true by default we should send the last 5 articles without warnings.

diegoquinteiro:
It turns out that defaulting boolean fields to true doesn't work properly. Instead I've replaced the flag with a new flag to allow articles with warnings - this one defaults to false now.

I've also added an UI to override this setting per article:
screen shot 2016-10-05 at 4 00 38 am

demoive:
I made a proposal to update the copy of the setting option in #490. Submitted as a separate PR on top of this branch since I don't have write access to the repo to add the commit directly to this PR. Feel free to update the branch with the same changes or merge my #490 PR after this one.

class-instant-articles-publisher.php
```diff
@@ -99,12 +101,14 @@ public static function submit_article( $post_id, $post ) {
 
 				// Don't process if contains warnings and blocker flag for transformation warnings is turned on.
 				if ( count( $adapter->transformer->getWarnings() ) > 0
-				  && isset( $publishing_settings['block_publish_with_warnings'] )
-					&& $publishing_settings['block_publish_with_warnings'] ) {
+				  && ( ( ! isset( $publishing_settings[ 'publish_with_warnings' ] ) ) || ! $publishing_settings[ 'publish_with_warnings' ] )
```
demoive: Nit: More consistent to remove the parenthesis enclosing the negated `isset` function?

```
   +---------------------------------------------------------------------------------------------------------------------+
   | +------------------------- remove? --------------------------+                                                      |
   | |                                                            |                                                      |
   v v                                                            v                                                      v
&& ( ( ! isset( $publishing_settings[ 'publish_with_warnings' ] ) ) || ! $publishing_settings[ 'publish_with_warnings' ] )
```

philipjohn: Yeah, probably no need for those :)

js/instant-articles-meta-box.js
```diff
@@ -1,3 +1,13 @@
+function instant_articles_force_submit ( post_ID ) {
+	var data = {
+		'action': 'instant_articles_force_submit',
+		'post_ID': post_ID,
+		'force': jQuery( '#instant_articles_force_submit' ).is( ':checked' )
```
philipjohn: As this is an admin action that updates the database we should be setting and [verifying a nonce](https://codex.wordpress.org/Function_Reference/check_ajax_referer).

meta-box/meta-box-template.php
```diff
@@ -163,6 +163,13 @@
 			</li>
 		<?php endforeach; ?>
 	</ul>
+	<?php if ( ! $publish_with_warnings ) : ?>
+		<hr />
+		<p>
+			<input type="checkbox" id="instant_articles_force_submit" <?php if ( $force_submit ) : ?>checked<?php endif; ?> />
```
philipjohn: WP's `checked()` function should be used here rather than printing out 'checked'

class-instant-articles-publisher.php
diegoquinteiro: nice catch! =) addressed in the latest commit

diegoquinteiro:
I just realized that this will potentially block some articles from being sent on the batch submission we do when submitting the page for review.

I'll address this properly.

diegoquinteiro:
Done!

everton-rosario:
Looks good to me!

PR #450, opened 2016-09-11
Do not block page selection if URL is not claimed

diegoquinteiro:
This is blocking the wizard on integrations for sites where the root URL is overwritten by server rules.


everton-rosario:
LGTM.
@philipjohn please include this on the next release.

PR #455, opened 2016-09-12
Only set https src on scripts that have a src attribute

### LINKED ISSUE #447: The relative-to-absolute script conversion process breaks inline scripts
I've noticed that the loop [here](https://github.com/Automattic/facebook-instant-articles-wp/blob/master/class-instant-articles-post.php#L698-L709) which iterates over all scripts elements in the content is setting a `src` attribute on all script tags, not just those that already have a src.

This means that inline scripts which are intended to run in interactive or ad iframes, for example, will not be executed in the final Instant Article.

As one example, I'm inserting ads in article content and using a custom transformer rule to process them into `.op-ad` elements. The ad markup looks roughly like this:

```
<figure class="op-ad">
  <iframe width="300" height="250">		
    <script src="https://www.googletagservices.com/tag/js/gpt.js"></script>
    <div id="js-gpt-m-instant-article-rec-fbia">
      <script type="text/javascript">
        googletag.pubads().enableSyncRendering();
        googletag.pubads().display(
          "/ad/slot/identifier",
          [300,250],
          "js-gpt-m-instant-article-rec-fbia"
        );
      </script>
    </div>
  </iframe>
</figure>
```

However, after going through the transformers, it ends up like this:

```
<figure class="op-ad">
  <iframe width="300" height="250">		
    <script src="https://www.googletagservices.com/tag/js/gpt.js"></script>
    <div id="js-gpt-m-instant-article-rec-fbia">
      <script type="text/javascript" src="https://">  // <---- See this unneeded src attribute?
        googletag.pubads().enableSyncRendering();
        googletag.pubads().display(
          "/ad/slot/identifier",
          [300,250],
          "js-gpt-m-instant-article-rec-fbia"
        );
      </script>
    </div>
  </iframe>
</figure>
```

It would be better to make sure that the script tags have a src attribute in the first place before setting it.

goldenapples:
Fixes #447

Per the HTML5 spec, [if the src attribute is present on a script tag, the inner content must be empty](https://www.w3.org/TR/html5/scripting-1.html#the-script-element).

Setting all scripts to an absolute url using https protocol breaks script tags where the script is inline. (when a script tag This change checks to makes sure that the src attribute is not empty before updating it to an absolute url.

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -704,6 +704,7 @@ class Instant_Articles_Post {
 					$explode_src = parse_url( $src );
 					if ( is_array( $explode_src ) && empty( $explode_src['scheme'] ) && ! empty( $explode_src['host'] ) && ! empty( $explode_src['path'] ) ) {
 						$src = 'https://' . $explode_src['host'] . $explode_src['path'];
+						$src = esc_url( $src );
 					}
 					$script->setAttribute( 'src' , $src );
 				}
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -700,11 +700,13 @@ class Instant_Articles_Post {
 		if ( ! empty( $scripts ) ) {
 			foreach ( $scripts as $script ){
 				$src = $script->getAttribute( 'src' );
-				$explode_src = parse_url( $src );
-				if ( is_array( $explode_src ) && empty( $explode_src['scheme'] ) && ! empty( $explode_src['host'] ) && ! empty( $explode_src['path'] ) ) {
-					$src = 'https://' . $explode_src['host'] . $explode_src['path'];
+				if ( ! empty( $src ) ) {
+					$explode_src = parse_url( $src );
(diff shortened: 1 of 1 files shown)


class-instant-articles-post.php
```diff
@@ -700,11 +700,13 @@ public function to_instant_article() {
 		if ( ! empty( $scripts ) ) {
 			foreach ( $scripts as $script ){
 				$src = $script->getAttribute( 'src' );
-				$explode_src = parse_url( $src );
-				if ( is_array( $explode_src ) && empty( $explode_src['scheme'] ) && ! empty( $explode_src['host'] ) && ! empty( $explode_src['path'] ) ) {
-					$src = 'https://' . $explode_src['host'] . $explode_src['path'];
+				if ( ! empty( $src ) ) {
+					$explode_src = parse_url( $src );
+					if ( is_array( $explode_src ) && empty( $explode_src['scheme'] ) && ! empty( $explode_src['host'] ) && ! empty( $explode_src['path'] ) ) {
+						$src = 'https://' . $explode_src['host'] . $explode_src['path'];
```
philipjohn: We should probably escape URL (using `esc_url()`)

everton-rosario: @philipjohn I've added the esc_url function inline here.

Issue #337, opened 2016-06-23
Embed filters affecting content outside of Instant Articles feeds


paulschreiber:
The FBIA plugin adds several filters. Example: in `themes/vip/plugins/facebook-instant-articles-2.11/embeds.php`, on line 54:

```
add_filter( 'embed_oembed_html', 'instant_articles_embed_oembed_html', 10, 4 );
```

In the function instant_articles_embed_get_html(), the HTML gets wrapped in a div like so: 

```
$html = sprintf( '<div class="embed">%s</div>', $html );
```

This code currently affects content in regular web views. It should only affect content in instant articles feeds.

kevinlisota:
I'm seeing this as well. The plugin's filters should not alter oembed handling or filters for front-end display on WP, and it shouldn't be inserting content there either. Need to restrict this oembed handling to plugin-related API calls or feeds only.

everton-rosario:
@paulschreiber @kevinlisota Are you guys experiencing malfunctioning because of the filters?

We need this so the transformer can have any oembed content transformed int "op-interactive".

Do you guys have any fix for this?

Can you provide more info/debuggable content so we can reproduce the same issues you are facing? So we can try another solution for the oembeds.

Thanks for your inputs. 
@paulschreiber sorry for the long time to act on this issue.

kevinlisota:
Reproducing this is easy. Create an article with an oembed(i.e. YouTube video or something). Turn on the plugin and view the page source for that article on WP front-end. Now turn off the plugin and view the page source again. The oembed will be wrapped with `<div class="embed">`. Essentially your plugin is hooking in to the `embed_oembed_html` filter wherever it is used, not just where it is needed which is in the context of your plugin.

I think you should be able to add the `embed_oembed_html` filter right before you need it, then immediately remove it so that it doesn't hook into other uses. I think (but haven't tested), that it might be right before `the_content` filter call that you make in class-instant-articles-post.php, but I'd have to dig in and see where the oembed filter is actually firing in WP.

I actually have a similar concern with the call to `remove_all_filters( 'embed_oembed_html' )` in embeds.php. From the look of it, I think that the moment the plugin loads, it removes anyone else who is hooked to `embed_oembed_html`, potentially affecting the front-end WP content in addition to what your generator does. Once again, this should only be called when you are working with IA content, not the HTML generated for the WP front-end.

[...]

PR #457, opened 2016-09-13
Don't try to load FB embeds compat unless we're in Instant Articles context

rinatkhaziev:
See #337 

Here's a different approach to #401 PR for that ticket,  this is just for fb embeds, I haven't tested it for the rest of the layers. 

The idea here is that 'instant_articles_before_transform' post runs only for IA context, so we can relatively safely attach our stuff here and it won't affect the rest of contexts.

Small extra - `op-social` was deprecated awhile back, so I replaced that with `op-interactive`

### DIFF
--- a/compat/class-instant-articles-jetpack.php
+++ b/compat/class-instant-articles-jetpack.php
@@ -39,6 +39,11 @@ class Instant_Articles_Jetpack {
 	 */
 	private function _fix_facebook_embed() {
 
+		// Don't try to fix facebook embeds unless we're in Instant Articles context
+		// to prevent mangled output on frontend
+		if ( !doing_action( 'instant_articles_before_transform_post' ) )
+			return;
+
 		// All of these are registered in jetpack/modules/shortcodes/facebook.php
 
 		if ( defined( 'JETPACK_FACEBOOK_EMBED_REGEX' ) ) {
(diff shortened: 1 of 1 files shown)


everton-rosario:
@diegoquinteiro can you double check this?

@rinatkhaziev Thanks for your PR!

philipjohn:
@diegoquinteiro @everton-rosario Any chance of getting this through soon folks?

compat/class-instant-articles-jetpack.php
```diff
@@ -39,6 +39,11 @@ private function _fix_youtube_embed() {
 	 */
 	private function _fix_facebook_embed() {
 
+		// Don't try to fix facebook embeds unless we're in Instant Articles context
+		// to prevent mangled output on frontend
+		if ( !doing_action( 'instant_articles_before_transform_post' ) )
+			return;
```
dashaluna: I'm not sure how to do it via github, but there are some Coding Standards that need fixing:

```
// Don't try to fix facebook embeds unless we're in Instant Articles context.
// This prevents mangled output on frontend.
if ( ! doing_action( 'instant_articles_before_transform_post' ) ) {
    return;
}
```

diegoquinteiro:
I've been struggling to find a good solution for this, but now #568 found a very simple way. In fact the `doing_action` hook wouldn't work as the embed processing code actually executes outside of the '.._before_..' action, actually in between the "before" and "after" actions.

@rinatkhaziev can we refactor this on top of #568 solution?

rinatkhaziev:
@diegoquinteiro makes sense, i'm going to try to carve out some time this week to do so, I have a comment on #568 which I'll post there in a sec.

diegoquinteiro:
@rinatkhaziev  thanks a lot! =)

diegoquinteiro:
I've actually went ahead and made the changes myself =D Thanks a lot @rinatkhaziev!

rinatkhaziev:
Thanks Diego! Sorry for dropping the ball

PR #464, opened 2016-09-16
Fixes #452 - Use of empty() on PHP prior to v5.5

### LINKED ISSUE #452: Plugin version 3.1.2 has error (wizard-template.php on line 101)
I just updated the plugin to 3.1.2 (also deleted and installed the plugin), and I see the following error when I go to "Instant Articles" in Dashboard:

```
Fatal error: Can't use function return value in write context in /home/myusername/public_html/wp-content/plugins/fb-instant-articles/wizard/templates/wizard-template.php on line 101
```

demoive:
This PR:

* [x] Makes use of `count()` instead of `empty()` when checking for [`get_settings_errors()`](https://codex.wordpress.org/Function_Reference/get_settings_errors) since it always returns an Array.

Fixes #452

### DIFF
--- a/wizard/templates/wizard-template.php
+++ b/wizard/templates/wizard-template.php
@@ -98,7 +98,7 @@
 
 <?php if ( ! $ajax ) : ?>
 	</div>
-	<?php if ( ! empty( get_settings_errors() ) ) : ?>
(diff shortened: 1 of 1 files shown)


everton-rosario:
LGTM.

@philipjohn can you have it released on the patch fix?

PR #497, opened 2016-10-11
Getter was renamed on SDK, updating here too

diegoquinteiro:
This fixes the name of the getter to the correct one on the SDK.


diegoquinteiro:
@philipjohn it is very important that this is merged before launching 3.2 tomorrow. This silly naming confusion will potentially breaks all article transformations.

Issue #275, opened 2016-05-18
Check box to disable Instant Articles on a Post


noumaan:
It would be nice to have options that allow site owners to choose which articles they want to appear in their Instant Articles feed. A checkbox in the 'Facebook Instant Articles' metabox below the post editor would be very useful. 

Another way is to allow site owners to choose a tag or category for articles they want to automatically add into Instant Articles feed.

katsar0v:
Yes, I would really love to have that, and guess it is not that complicated to build

katsar0v:
Any update on this feature? I guess it is a must for such a plugin

cuckoofordeals:
Yes I would love this as well!

donaldG:
@noumaan @katsar0v @cuckoofordeals I was able to achieve this by creating a category specifically for the Instant Articles. Figuring out that feed URL was a bit of a pain. What I figured out was after creating the Category my feed was this:
http://www.site.com/category/fb-instant-articles-your-custom-cat-name/feed/instant-articles

The documentation isn't super clear on that. Just after trial & error I discovered it. This is also without setting a custom INSTANT_ARTICLES_SLUG. I think all that would do is change 'instant-articles' above to 'your-custom-instant-articles-slug' but everything else would be the same.

stuartshields:
@donaldG You don't need /feed/ in the URL (well I haven't come across any instances of needing this).

donaldG:
Woah, you're right. I couldn't get it working any other way but you're
totally right. In addition, /feed/ does work, for whatever that's worth.

katsar0v:
Somehow posts still get published, though they are basically not in the feed:

https://github.com/Automattic/facebook-instant-articles-wp/issues/460

katsar0v:
Still no updates, but I see this as a major issue

volokolamskspb:
Hi, I fixed the plugin. Now there's a checkbox which filters unwanted posts from RSS Feed.
I know it's a bit ugly solution: I added a new table with ID of posts which will be displayed in the feed.
Just install updated the plugin: https://github.com/volokolamskspb/FB_Instant_Articles_wp/
Thanks

daffsandcomp:
can someone please explain to me in the most simple way possible if there is a way of turning of instant articles for certain posts?

PR #504, opened 2016-10-17
Add instant_articles_should_submit_post filter hook  to allow developers control of whether the post should be submitted to IA

### LINKED ISSUE #503: Ability to disable API publishing/feed publishing via filter
Currently there's no way to arbitrarily disable post submission to Instant Articles either via API or feeds. This is a pretty serious flaw in workflow, people definitely want the ability to have control over that. The first step would be to add a filter hook so that theme/plugin developers could build on top of that. 

Proposed solution is to add `instant_articles_should_submit_post` filter that defaults to `true` (PR incoming).

See #442 #275 #186

rinatkhaziev:
This PR:

adds `instant_articles_should_submit_post` filter hook to allow developers to control whether current post should be submitted to Instant Articles or not. 

The filter defaults to `TRUE` - the  post should be submitted to IA. We pass instance of Instant_Articles_Post to the filter.

Relates to #503 442 #275 #186

Fixes #503

### DIFF
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -57,6 +57,11 @@ class Instant_Articles_Publisher {
 		// Transform the post to an Instant Article.
 		$adapter = new Instant_Articles_Post( $post );
 
+		// Allow to disable post submit via filter
+		if ( false === apply_filters( 'instant_articles_should_submit_post', true, $adapter ) ) {
+			return;
+		}
+
 		$article = $adapter->to_instant_article();
 
 		// Skip empty articles or articles missing title.
--- a/feed-template.php
+++ b/feed-template.php
@@ -21,6 +21,11 @@ $last_modified = null;
 			<?php
 			$instant_article_post = new Instant_Articles_Post( get_post( get_the_id() ) );
 
+			// Allow to disable post submit via filter
+			if ( false === apply_filters( 'instant_articles_should_submit_post', true, $instant_article_post ) ) {
+				continue;
+			}
+
 			// If we’re OK with a limited post set: Do not include posts with empty content -- FB will complain.
 			if ( defined( 'INSTANT_ARTICLES_LIMIT_POSTS' ) && INSTANT_ARTICLES_LIMIT_POSTS && ! strlen( trim( $instant_article_post->get_the_content() ) ) ) {
 				continue;


banago:
When is this planned to be merged?

rinatkhaziev:
@banago this issue tracked in #186 and i think it's scheduled for 3.3 but it's still not merged into master. /cc @everton-rosario

everton-rosario:
@rinatkhaziev Thanks for putting this PR out.
It will be helpful for lots of partners.

Your contributions makes Facebook Instant Articles great!

mjangda:
Also related: #74

mjangda:
Should we run this filter when rendering the metabox and adjust what's displayed there?

rinatkhaziev:
@mjangda probably it should be the other way around - we should add an UI element (ie checkbox) that, if checked will add a post meta to indicate this post should be skipped, and then run the filter on the value of meta field? 

Something along the lines of 
`
if ( false === apply_filters( 'instant_articles_should_submit_post', '' === get_post_meta( $post->ID, 'instant_articles_skip_post', true ) , $adapter ) ) {
		return;
	}`

osterkraft:
First of all, thanks for the good work on the plugin!

As I understand, next release will just include a notice if the post has been programmatically filtered out, is that correct? What is your recommended way to go about if we want to include a checkbox on the post page, as discussed earlier? 

[This seemingly abandoned and defunct plugin](https://sv.wordpress.org/plugins/instant-articles-filter/) ([small fix required](https://wordpress.org/support/topic/minor-fix-needed/)) provided a solution but it only filtered the feed, and had no effect on the API post process.

nders84:
anyone can enlighten me how to use this hook as a checkbox to select which articles to be published?

I would like the defauly set to FALSE, and when I check a checkbox  in the metabox and publish the post, it will get converted to Instant Articles.

Really need help on this. Thanks in advance!

volokolamskspb:
@nders84 
I need the same functionality. I'm trying to fix that problem by editing the plugin.
I don't understand how and where do I solve that problem using the hook.
I'll post my solution when I'm done.

amansarwar:
> @nders84
> I need the same functionality. I'm trying to fix that problem by editing the plugin.
> I don't understand how and where do I solve that problem using the hook.
> I'll post my solution when I'm done.

Did you figured this out yet?

HRarsha:
> @nders84 I need the same functionality. I'm trying to fix that problem by editing the plugin. I don't understand how and where do I solve that problem using the hook. I'll post my solution when I'm done.

Can you share the solution please? Isn't there any way to do this without touching the code? simply with a checkbox on post edit screen.

PR #522, opened 2016-11-04
Created a new Ingestion flow for Instant Articles based on Open Graph

everton-rosario:
This PR:

* [x] Uses flow to new OpenGraph ingestion
* [x] Simplifies the configuration on the WordPress plugin, since only page ID is needed 
* [x] Keeps retro-compatibility with API setup
* [x] Creates a switch to enable API or OPEN GRAPH ingestion

This ingestion flow takes in consideration how is the current status of the setup.

 If setup is already complete on API, it means a migration,
   set the configuration as API and the flow goes that way
 otherwise
   set the configuration as Open Graph and flow goes this way


everton-rosario:
This PR wont be merged.
This is the replacing PR: https://github.com/Automattic/facebook-instant-articles-wp/pull/548

PR #531, opened 2016-11-14
Dutch Translations

xvilo:
This PR:

Adds Dutch Translations from the POT file.


xvilo:
So I saw there is much to do regarding the translations, I will do those first and make a new PR!

PR #535, opened 2016-11-16
Fixes #515 - Enables Playbuzz plugin out of the box

### LINKED ISSUE #515: Playbuzz integration with IA
PB Content can be embedded in two ways:
1. Code snippet (including Script element, that can reside in the head element and div with class="pb_feed") - Doesn't required pb plugin to be installed
2. Shortcodes - Requires pb plugin to be installed
# Issue 1

When pb plugin isn't installed - pb code snippet transformation is not activated. In most cases the code snippet method of embedding is used on cases pb plugin isn't installed resulting in IA pages not including the pb embedded articles.
# Expected Result

PB transformation should be part of the default transformation rules set and not depending on the pb plugin existence. 
# Steps required to reproduce the issue

Embed a playbuzz code snippet without pb plugin installed

Snippet:

```
<script type="text/javascript" src="//cdn.playbuzz.com/widget/feed.js"></script>
<div class="pb_feed" data-embed-by="7cab61df-eb6a-41ae-bd54-8ab76018c04c" data-game="/thesun18/guess-the-celebrity-behind-the-halloween-costume"></div>

```

Selector should be based ONLY on the div.pb_feed (script should be ignored) and transformed into:

```
      <figure class="op-interactive">
        <iframe class="no-margin">
          <script type="text/javascript" src="//cdn.playbuzz.com/widget/feed.js"></script>
          <div class="pb_feed" data-embed-by="7cab61df-eb6a-41ae-bd54-8ab76018c04c" data-game="/thesun18/guess-the-celebrity-behind-the-halloween-costume"></div>
        </iframe>
      </figure>

```

---
# Issue 2

Playbuzz items being embedded with short codes are not being transformed
# Expected Result

To transform short code the same way the raw code snippet works
# Steps required to reproduce Issue

Install PB plugin.
Embed an item using playbuzz plugin (which uses shortcodes)
Published item doesn't contain pb embedded article

*\* Version Info **
- Plugin version: 3.2
- WordPress version: 4.6.1
- PHP version: 5.5.9-1ubuntu4.20

everton-rosario:
This PR:

* [x] Removed conditional to Playbuzz plugin, so it makes available for all users
Fixes #515

### DIFF
--- a/tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
+++ b/tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
@@ -73,7 +73,6 @@ class WPTransformerTest extends \PHPUnit_Framework_TestCase
 
         $this->assertEquals($expected, $result);
         // there must be 3 warnings related to <img> inside <li> that is not supported by IA
-        // And 1 warning related to the getter
         $this->assertEquals(3, count($transformer->getWarnings()));
     }
 
--- a/tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
+++ b/tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
@@ -6,7 +6,7 @@
  * This source code is licensed under the license found in the
  * LICENSE file in the root directory of this source tree.
  */
-namespace Facebook\InstantArticles\Transformer\CMS;
+namespace Facebook\InstantArticles\Transformer;
 
 use Facebook\InstantArticles\Transformer\Transformer;
 use Facebook\InstantArticles\Elements\InstantArticle;
@@ -74,7 +74,7 @@ class WPTransformerTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals($expected, $result);
         // there must be 3 warnings related to <img> inside <li> that is not supported by IA
         // And 1 warning related to the getter
-        $this->assertEquals(4, count($transformer->getWarnings()));
+        $this->assertEquals(3, count($transformer->getWarnings()));
     }
 
     public function testTitleTransformedWithBold()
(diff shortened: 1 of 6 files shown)


diegoquinteiro:
Why would you remove the conditional? We don't need to include the code if they don't have the playbuzz plugin installed, do we?

everton-rosario:
For release Thursday/Friday. Dec 15/16

everton-rosario:
@diegoquinteiro This will be add for any one out of the box, since Playbuzz noticed that most of users copy/paste the snippet of code having or not the playbuzz plugin.

tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
```diff
@@ -6,7 +6,7 @@
  * This source code is licensed under the license found in the
  * LICENSE file in the root directory of this source tree.
  */
-namespace Facebook\InstantArticles\Transformer\CMS;
```
diegoquinteiro: Folder structure should match the namespace for convenience. I don't see the need to change it here, btw.

tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
```diff
@@ -74,7 +74,7 @@ public function testTransformerLikeWPContent()
         $this->assertEquals($expected, $result);
         // there must be 3 warnings related to <img> inside <li> that is not supported by IA
         // And 1 warning related to the getter
-        $this->assertEquals(4, count($transformer->getWarnings()));
+        $this->assertEquals(3, count($transformer->getWarnings()));
```
diegoquinteiro: Please update the comment above to reflect the new expectation.

tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
everton-rosario: This was an old structure when it was migrated from SDK.
Here it should match the folder structure.

tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
everton-rosario: Nice catch @diegoquinteiro.

Fixed the comments. Since the tests started using the fragment, and the rules were better structured, no getter exception was raised.

PR #548, opened 2016-11-29
Og ingestion clean

everton-rosario:
This PR:
* [x] Creates a new Ingestion flow for instant articles
* [x] Simplifies the configuration on the WordPress plugin, since only page ID is needed 
* [x] Removed the configuration wizard, since the new ingestion flow covers all features we had on the API


facebook-instant-articles.php
```diff
@@ -321,20 +314,49 @@ function inject_url_claiming_meta_tag() {
...
+	 *
+	 * @since 4.0
+	 */
+	function inject_op_markup_meta_tag() {
+		$post = get_post();
+		// Transform the post to an Instant Article.
+		$adapter = new Instant_Articles_Post( $post );
+		$url = $adapter->get_canonical_url();
+		?>
+		<meta property="op:markup_url" content="<?php echo esc_attr( $url ); ?>?op=1" />
```
demoive: @everton-rosario what's the `op` parameter used for?

We should use the [`add_query_arg()`](https://developer.wordpress.org/reference/functions/add_query_arg/) function here:

```
esc_attr( add_query_arg( array( 'op' => '1' ), $url ) );
```

everton-rosario: Nice catch!

diegoquinteiro:
Moved to #643

PR #562, opened 2016-12-16
Add release script

diegoquinteiro:
Created a script to automatically:

- [x] Bump the version of the plugin in the right files
- [x] Create a tag for the version
- [x] Commit and tag the repository with the version
- [x] Create a new release for the version
- [x] Uploads the binary zip for the plugin
- [x] Release on WordPress.org

Basic usage: ./release.sh -v <VERSION>

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -12,15 +12,20 @@ function show_help {
 cat <<EOF
 
 Usage:
-  release.sh [-hvs] [version]
+  release.sh [-hvs] [-c <command>] [version]
 
 Arguments:
-  version - The target version (ex: 3.2.1)
+  release      - Creates a GitHub released based on existing version tag
+  bump_version - Creates a new version tag
+  version      - The target version (ex: 3.2.1)
 
 Options:
-  -h Display this help message
-  -v Verbose mode
-  -s Simulate only (do not release)
+  -h            Display this help message
+  -v            Verbose mode
+  -s            Simulate only (do not release)
+  -c <command>  Runs only a single command. Possible commands are:
+                  - bump_version: generate a new version tag
+                  - release: release a new version
 
 EOF
 }
@@ -66,8 +71,9 @@ OPTIND=1
 # Read options:
 verbose=0
 simulate=0
+selected_cmd='all'
 
-while getopts "hvs" opt; do
+while getopts "hvc:s" opt; do
     case "$opt" in
     h|\?)
         show_help
@@ -76,6 +82,8 @@ while getopts "hvs" opt; do
     v)  verbose=1
         ;;
     s)  simulate=1
+        ;;
+    c)  selected_cmd="${OPTARG}"
         ;;
     esac
 done
@@ -110,6 +118,12 @@ else
   message "Found git: $(git --version)"
 fi
 
+if ! type "js" > /dev/null; then
+  error_message "SpiderMonkey interpreter not found, please install SpiderMonkey before continuing"
+else
+  message "Found SpiderMonkey"
+fi
+
 #------------------------------------
 # Check if we are in the right folder
 #------------------------------------
@@ -140,9 +154,6 @@ function revert_repo {
     message "Applying stashed changes"
     run git stash apply $stash_ref
   fi
-  if [[ -e resty ]]; then
-    rm resty
-  fi
 }
 
 function confirm {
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -0,0 +1,330 @@
...
+reset=`tput sgr0`
+
+#--------------------------------
+# Functions
+#--------------------------------
+function show_help {
+cat <<EOF
+
+Usage:
+  release.sh [-hvs] [-c <command>] [version]
```
everton-rosario: Add here the top variable that contains the name of file.

release.sh
```diff
@@ -0,0 +1,330 @@
+#!/bin/bash
+red=`tput setaf 1`
+green=`tput setaf 2`
+yellow=`tput setaf 3`
+blue=`tput setaf 4`
+reset=`tput sgr0`
+
```
everton-rosario: If possible, try to use the `ps` function to discover the command line name, so this way this will have the name not hardcoded into script.

`basename` is the function you might be looking for: http://stackoverflow.com/questions/192319/how-do-i-know-the-script-file-name-in-a-bash-script

release.sh
```diff
@@ -0,0 +1,330 @@
...
+  version      - The target version (ex: 3.2.1)
+
+Options:
+  -h            Display this help message
+  -v            Verbose mode
+  -s            Simulate only (do not release)
+  -c <command>  Runs only a single command. Possible commands are:
+                  - bump_version: generate a new version tag
+                  - release: release a new version
+
```
everton-rosario: Consider adding example for fast track the calls:
- Release complete
- Try the release
- Just bump version
- Just release
And cases you consider the most frequently used

release.sh
```diff
@@ -0,0 +1,330 @@
...
+EOF
+}
+
+function invalid_usage {
+  printf $red
+  echo $@
+  echo "Aborting..."
```
everton-rosario: The script is really fast, so no need for the `...`
I guess `"Execution Aborted"` is enough.

release.sh
```diff
@@ -0,0 +1,330 @@
...
+  printf $reset
+  show_help
+  exit -1;
+}
+function error_message {
+  printf $red
+  echo $@
+  echo "Aborting..."
```
everton-rosario: The script is really fast, so no need for the `...`
I guess `"Execution Aborted"` is enough.

release.sh
```diff
@@ -0,0 +1,464 @@
...
+function run {
+  if [[ $simulate == 1 ]]; then
+    run_message $@
+  else
+    "$@"
+  fi
+}
+
+function revert_repo {
+  run cd $repo_dir
```
akeem: I am not sure but I think this might halt if there are uncommitted files in the working directory when this method is called

release.sh
```diff
@@ -0,0 +1,464 @@
...
+    message "Switching to master..."
+    run git checkout master
+  fi
+
+  message "Pulling latest version from GitHub"
+  run git pull --rebase
+
+  if [[ ! -e resty ]]; then
+    message "Downloading resty to connect to GitHub API"
+    run curl -L http://github.com/micha/resty/raw/master/resty > resty
```
akeem: For this and jsawk we should consider pulling from a fixed tag/releases, hopefully this will avoid scenarios where someone is burned by broken master branches

release.sh
```diff
@@ -0,0 +1,464 @@
...
+
+  message "Uploading binary for release..."
+
+  if [[ github_access_token ]]; then
+    response=$(run curl -H "Authorization: token $github_access_token" -H "Content-Type: application/zip" --data-binary @facebook-instant-articles-wp.zip $upload_url\?name=facebook-instant-articles-wp-$version.zip )
+  else
+    response=$(run curl -u $github_username:$github_password -H "Content-Type: application/zip" --data-binary @facebook-instant-articles-wp.zip $upload_url\?name=facebook-instant-articles-wp-$version.zip )
+  fi
+
+  run rm facebook-instant-articles-wp.zip
```
akeem: How would you feel about adding a check around this deletion? We may be able to use the status code from the curl request if we pass a few extra parameters to the call

release.sh
```diff
@@ -0,0 +1,464 @@
...
+  composer install
+
+  message "Checking out SVN repository..."
+  run cd $tmp_dir
+  run svn checkout -q https://plugins.svn.wordpress.org/fb-instant-articles/
+  run cd fb-instant-articles/trunk/
+
+  confirm "Copy new version to trunk?"
+  run cp -rf $repo_dir/* ./
+  run rm -rf .[!.]*
```
akeem: I lame, what would this regex match against?

akeem:
Over all, looks good. I left a few nit comments

release.sh
diegoquinteiro: For simplicity I wouldn't keep the zip file even if the upload wasn't successful. Reason being I want it to always create the file again in order to avoid dealing with the invalidation of the file (due to changes on the repo).

release.sh
diegoquinteiro: That's removing all hidden files (as they're not code files and thus not used in the SVN release). I'll add a comment to the code too.

release.sh
diegoquinteiro: I'll add a git stash for safety and comment on the code.

release.sh
diegoquinteiro: Good call =)

Issue #426, opened 2016-08-31
Add a conditional or filter which can be checked to process content only for instant articles feed


goldenapples:
I've been trying to attach some filters to the_content on Instant Articles only. Currently I'm doing this by checking `is_feed( INSTANT_ARTICLES_SLUG )` in my filter function.

However, in order to do this, I've had to disable the Instant Articles Publisher submit_article function, as the condition isn't true then, and short of doing a debug backtrace, there isn't a straightforward way of knowing at that point whether the content filter is being run for an instant article or some other view.

It would be very nice if there was either:
- a separate filter registered, something like `instant_articles_content`, which is processed before `the_content` in `Instant_Articles_Post::get_the_content()`, or
- a conditional function like `is_instant_article()`; which returns true when called from a filter on an instant article and false otherwise. 

Having either one of these in place would enable processing instant article content before transformation, which would be helpful in many circumstances.

diegoquinteiro:
There is now a filter `instant_article_content`. See: https://github.com/Automattic/facebook-instant-articles-wp/blob/master/class-instant-articles-post.php#L349

goldenapples:
Thanks! Unfortunately, that doesn't work for my use cases, as I need something that can either run before 'the_content', or disable some of the normal filters on the_content when in Instant Article context.

Cases in point: 

* I have shortcodes which need to output different markup in Instant Articles than they do in web views. There's currently no way of doing this other than to disable the Instant Article publisher and the meta box on the post, because otherwise there's no way to determine whether the content filter is being run for an instant article or not.

* I want to be able to split the content before wpautop in order to insert ads or related article links at certain points in the content, and these rules and the markup inserted will be different on Facebook than on the web. Currently there's no way to attach a filter to Instant Article content before wpautop, because it's forcibly added before processing the content even if you disable it.

I'll prepare a pull request with the idea I had for this - probably creating an `is_instant_article()` conditional which can be checked and will return true if processing an article for a feed or for the Publisher.

PR #568, opened 2016-12-22
Add `is_transforming_instant_article()` conditional

goldenapples:
Adds a new function in the global namespace to check if the current view
being processed is for an instant article. Handles both the RSS feed and
content which is being processed for the Instant_Articles_Publisher
class. This conditional can be checked in any filters on 'the_content',
'the_title' or elsewhere that need to treat Instant Articles content
separately from web content.

See #426

### DIFF
--- a/facebook-instant-articles.php
+++ b/facebook-instant-articles.php
@@ -151,12 +151,29 @@ if ( version_compare( PHP_VERSION, '5.4', '<' ) ) {
 	 * @since 0.1
 	 */
 	function instant_articles_feed() {
+		is_instant_article( true );
 
 		// Load the feed template.
 		include( dirname( __FILE__ ) . '/feed-template.php' );
 
+		is_instant_article( false );
 	}
 
+	/**
+	 * Whether currently processing an instant article.
+	 *
+	 * @param bool Set the status
+	 * @return bool
+	 */
+	function is_instant_article( $set_status = null ) {
+		static $is_instant_article = false;
+
+		if ( isset( $set_status ) ) {
+			$is_instant_article = (bool) $set_status;
+		}
+
+		return $is_instant_article;
+	}
 
 	/**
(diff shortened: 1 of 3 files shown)


goldenapples:
I know that this logic could easily be done in a theme or another plugin that interfaces with this, now that the two actions are available, but since this has been a common need of mine, it would be nice to have something like this available directly in this plugin itself.

facebook-instant-articles.php
```diff
@@ -157,6 +157,23 @@ function instant_articles_feed() {
 
 	}
 
+	/**
+	 * Whether currently processing an instant article.
+	 *
+	 * @param bool Set the status
+	 * @return bool
+	 */
+	function is_instant_article( $set_status = null ) {
```
diegoquinteiro: what about `is_transforming_instant_article`?

rinatkhaziev:
@goldenapples in f81998e you attach `is_instant_article` to `before..` and `after...`, this is potentially a point of breakage since hooks can be manipulated outside the context of the plugin. I feel like it's safer to put calls inside `Instant_Articles_Post::to_instant_article`  directly. 

cc @diegoquinteiro

facebook-instant-articles.php
goldenapples: Good sugeestion. I've updated the method name in aa33ecf.

goldenapples:
@rinatkhaziev Good call. I was trying to port over the ad hoc solution I was using in a theme to the plugin. But if this method is going into the plugin directly, it does make much more sense to set the status directly rather than plugging in to those hooks. Does ca03891 look like what you had in mind?

rinatkhaziev:
@diegoquinteiro This looks good to me, should we merge it to master and then I'm going to fix my PR?

diegoquinteiro:
@rinatkhaziev done! =)

PR #577, opened 2016-12-31
Import drafts as unpublished

### LINKED ISSUE #576: Submit drafts as unpublished to instant articles
*As a user, I want to be able to preview the generated instant article before my post is published to my website.*

Therefore, it would be nice if drafts would be submitted as unpublished to instant articles.

jppgks:
- [x] Import drafts as unpublished to instant articles
- [ ] ~Add option to settings to enable/disable the import of drafts~ Add filter

Closes #576

### DIFF
--- a/class-instant-articles-publisher.php
+++ b/class-instant-articles-publisher.php
@@ -43,8 +43,8 @@ class Instant_Articles_Publisher {
 			return;
 		}
 
-		// Don't process if this post is not published
-		if ( 'publish' !== $post->post_status ) {
+		// Don't process if this post is not published nor a draft
+		if ( 'publish' !== $post->post_status && 'draft' !== $post->post_status ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


diegoquinteiro:
Actually, considering #151, we should make the default to be `false` due to the feedback we got.

jppgks:
@diegoquinteiro That's the reason I suggested some kind of toggle in the settings 😉

jppgks:
Although a toggle in the settings might be too much? I don't know.
1. For a developer, it's no issue changing the behavior
2. A general user, though, might never access the 'import drafts' feature if it wouldn't be in the settings

Happy to hear your thoughts on this, @diegoquinteiro!

diegoquinteiro:
@jppgks makes sense, I'm with you. The UI can be a little harder though, so you could still go with putting the filter for now so we can land it faster, then you can create a new PR with the UI without holding you back from using the feature. What you think?

timjacobi:
@jppgks Are you still working on this?

timjacobi:
Closing due to inactivity. Will reopen if more info is posted.

PR #592, opened 2017-01-23
Add missing information to composer.json

onnimonni:
This way the plugin can be installed as dependency with composer.


diegoquinteiro:
Thanks a lot @onnimonni for your contribution!

PR #595, opened 2017-01-27
Add transformer rule for cite element

davebonds:
This PR:

Adds a transformer rule for the cite element adding the ItalicRule class.


diegoquinteiro:
@davebonds nice catch! Thanks a lot =)

PR #597, opened 2017-01-31
Don't try to load embeds compat unless we're in Instant Articles context

diegoquinteiro:
This PR is built on top of #457 and using the solution introduced by #568.

It prevents the compat layer from messing with the regular FE output, running only during the Instant Article transformation.

### DIFF
--- a/compat/class-instant-articles-jetpack.php
+++ b/compat/class-instant-articles-jetpack.php
@@ -39,10 +39,11 @@ class Instant_Articles_Jetpack {
 	 */
 	private function _fix_facebook_embed() {
 
-		// Don't try to fix facebook embeds unless we're in Instant Articles context
-		// to prevent mangled output on frontend
-		if ( !doing_action( 'instant_articles_before_transform_post' ) )
-			return;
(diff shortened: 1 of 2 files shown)


compat/class-instant-articles-jetpack.php
```diff
@@ -79,7 +85,7 @@ public static function facebook_embed_handler( $matches, $attr, $url ) {
 			$locale = 'en_US';
 		}
 
-		return '<figure class="op-social"><iframe><script src="https://connect.facebook.net/' . $locale . '/sdk.js#xfbml=1&amp;version=v2.6" async></script><div class="fb-post" data-href="' . esc_url( $url ) . '"></div></iframe></figure>';
+		return '<figure class="op-interactive"><iframe><script src="https://connect.facebook.net/' . $locale . '/sdk.js#xfbml=1&amp;version=v2.6" async></script><div class="fb-post" data-href="' . esc_url( $url ) . '"></div></iframe></figure>';
```
everton-rosario: Nice catch.

PR #604, opened 2017-02-06
Fix typo for $provider_url at L42, L44 and L46

abdusfauzi:
This PR:

Fixes #
few typos with variable spelling causing the added fix for Vimeo, Facebook and Vine not working.


diegoquinteiro:
Nice catch! Thank you!

PR #612, opened 2017-02-13
Add an option to set a footer

### LINKED ISSUE #261: Additional setting option to apply a global footer
This feature request is to add an additional option in the settings to set a footer which would apply to all articles.

Example use case: Useful for copyright information.

Reference:
https://wordpress.org/support/topic/footer-110?replies=1
https://wordpress.org/support/topic/footer-customization-4/
https://wordpress.org/support/topic/customizing-footer-5/

sakatam:
fixes #261

- [x] Set footer element during the construction of IA document
- [x] Rename "Style" setting to "Appearance"
- [x] Consolidate appearance settings logic into one function

image

image

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -678,16 +678,7 @@ class Instant_Articles_Post {
 				->addMetaProperty( 'op:generator:application', 'facebook-instant-articles-wp' )
 				->addMetaProperty( 'op:generator:application:version', IA_PLUGIN_VERSION );
 
-		$settings_style = Instant_Articles_Option_Styles::get_option_decoded();
-		if ( isset( $settings_style['article_style'] ) && ! empty( $settings_style['article_style'] ) ) {
-			$this->instant_article->withStyle( $settings_style['article_style'] );
-		} else {
-			$this->instant_article->withStyle( 'default' );
-		}
-
-		if ( isset( $settings_style['rtl_enabled'] ) ) {
-			$this->instant_article->enableRTL();
-		}
+		$this->set_appearance_from_settings();
 
 		$transformer->transformString( $this->instant_article, $this->get_the_content(), get_option( 'blog_charset' ) );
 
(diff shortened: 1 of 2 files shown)


diegoquinteiro:
There was an idea from @demoive and @everton-rosario to change the taxonomy and remove the high level "Footer" section.

I personally like the way @sakatam did it here for the simplicity and because it allows a nice future expansion to include other footer elements and options there (setting up Credits for instance).

Please @everton-rosario and @demoive add your reasoning to this thread so it is documented.

I'll approve the diff as it is and defer to @sakatam for the final decision on changing the taxonomy or not.

demoive:
Right, so from the requests I've seen, the reason to have the option for a global footer setting is to make it easy for setting the content of the footer globally. In this sense, I classify it as a setting for the "theme" of an article (or "template") which lends itself for being under the "Styles" section.

@everton-rosario made a point that if this were the case, perhaps "Styles" could be renamed to something else more encompassing, for example: 

- Appearance
- Template
- Styling
- IA markup setup
- Markup settings

demoive:
To @diegoquinteiro's point, however, in the future, I could also foresee the desire to define multiple footers, each to be applied to articles based on some condition — such as based on a category or tag. In this case, it could make sense to have a dedicated section for these settings, i.e. keep things as they are.

demoive:
Either way, I think this implementation is spot on. Great job here.

sakatam:
Thank you guys all for your suggestions! I agree with moving the copyright setting into the "Style" section and renaming the section to more encompassing one. Once we will need to expand the footer setting in the future, we can split it and have a dedicated section.

sakatam:
Pushed the new version. Now the copyright setting is in the new "Appearance" section which includes Style and RTL settings.

sakatam:
@diegoquinteiro @demoive @everton-rosario Do you guys have any remaining concerns on this change?

diegoquinteiro:
Looks great. Merging. =)

samichamoun:
Just updated to the latest version of the plugin and don't see this option??

sakatam:
@samichamoun Unfortunately, this one is not in v3.3. Will definitely be included in the next release!

samichamoun:
Ah, no worries. I notice that the screenshot says no tags can be added. Is there a reason why we can't put any hyperlinks in here as I've seen a number of pages using the footer for this?

rondavilar:
I've been inserting footers manually in every article which is tedious, plus they get replaced every time the article updates. Is there a way to do this at the PHP level with hyperlink tags? Completely new to programming.

sakatam:
@samichamoun 

> Is there a reason why we can't put any hyperlinks in here as I've seen a number of pages using the footer for this?

This is simply due to [the restriction of the SDK](https://github.com/facebook/facebook-instant-articles-sdk-php/blob/master/src/Facebook/InstantArticles/Elements/Footer.php#L105-L111) that this plugin depends on. However, since our platform allows hyperlinks in here, we should be able to do that within this plugin, too. I will file an issue for this. Thank you for pointing this out!

sakatam:
Filed an issue for rich formats: #619

PR #428, opened 2016-08-31
Refactor to use transform string

yssk22:
This PR:

* [x] improve encoding handling by IA the SDK function `transformString` and remove the complexities of `loadHTML` things.

Follows #395

Relates to #369, #344

Fixes #339

PR #624, opened 2017-02-24
fix merge damage

sakatam:
as @yssk22 commented in #428, the refactored code has been reverted back by mistake, and it result in garbled MB characters in title.

This PR:

* [x] puts back the code that has mistakenly been reverted

Relates to #428


sakatam:
Thanks @everton-rosario!

PR #632, opened 2017-03-06
Adds a default PassThroughRule for tag `<sup>`

### LINKED ISSUE #615: <sup> should be a common rule by default
Hi,

I think `<sup>` should be a common rule by default

everton-rosario:
This PR:

* [x] Adds PassThroughRule for tag `<sup>`

Fixes #615

### DIFF
--- a/rules-configuration.json
+++ b/rules-configuration.json
@@ -27,6 +27,9 @@
         "class": "ParagraphRule",
         "selector": "p"
     }, {
+        "class": "PassThroughRule",
+        "selector": "sup"
+    }, {
         "class": "LineBreakRule",
         "selector": "br"
     }, {


rules-configuration.json
```diff
@@ -27,6 +27,9 @@
         "class": "ParagraphRule",
         "selector": "p"
     }, {
+        "class": "PassThroughRule",
+        "selector": "sup"
```
vkama: Is there a reason that `<sub>` was not added too?

timjacobi:
@everton-rosario What work is left here to get this merged?

timjacobi:
Actually I'm not sure if I agree with this PR. I feel like we should rather define a new rule that converts a character wrapped in a `<sub>` tag to its superscript equivalent. Otherwise the widely used m² would be converted to m2 for example. I think that would make people unhappy.

paulschreiber:
@timjacobi can you propose a patch that does what you want?

jerclarke:
Someone at facebook, please point out how absurd it is that these simple HTML tags don't work in IA. 

It's sad that we have to consider these options.

everton-rosario:
Somehow this got lost in between releases.

I'll update the PR with the proper tags added.

everton-rosario:
We do render it not so gracefully into the UI of Instant Articles, though we do support the `<sub>`/`<sup>` tags.

I'll possibly change this PR to properly have the tags and the proper rules, so we don't need to bypass them, rendering it properly as `sup` and `sub` tags.

PR #643, opened 2017-03-16
Open Graph Ingestion Flow

diegoquinteiro:
This PR is a clean/rebased version of #548 

- [x] Creates a new Ingestion flow for instant articles
- [x]  Simplifies the configuration on the WordPress plugin, since only page ID is needed
- [x]  Removed the configuration wizard, since the new ingestion flow covers all features we had on the API


facebook-instant-articles.php
```diff
@@ -342,20 +334,47 @@ function inject_url_claiming_meta_tag() {
 		$publishing_settings = Instant_Articles_Option_Publishing::get_option_decoded();
 		$fb_page_settings = Instant_Articles_Option_FB_Page::get_option_decoded();
 
-		if ( isset( $fb_page_settings['page_id'] ) ) {
+		if ( isset( $fb_page_settings[ 'page_id' ] ) ) {
```
everton-rosario: nice catch 💅

facebook-instant-articles.php
```diff
@@ -342,20 +334,47 @@ function inject_url_claiming_meta_tag() {
...
 			?>
-			<meta property="fb:pages" content="<?php echo esc_attr( $fb_page_settings['page_id'] ); ?>" />
+			<meta property="fb:pages" content="<?php echo esc_attr( $fb_page_settings[ 'page_id' ] ); ?>" />
```
everton-rosario: nice catch 💅

facebook-instant-articles.php
```diff
@@ -342,20 +334,47 @@ function inject_url_claiming_meta_tag() {
...
+	 * Automatically add meta tag for Instant Articles Open Publish scraper
+	 *
+	 * @since 4.0
+	 */
+	function inject_op_markup_meta_tag() {
+		$post = get_post();
+		// Transform the post to an Instant Article.
+		$adapter = new Instant_Articles_Post( $post );
+		$url = $adapter->get_canonical_url();
+		$url = add_query_arg( 'op', '1', $url );
```
everton-rosario: Since we removed the op: ("Open publisher") thing. Shouldn't we change this op=1 thing?

facebook-instant-articles.php
```diff
@@ -342,20 +334,47 @@ function inject_url_claiming_meta_tag() {
...
+		?>
+		<meta property="ia:markup_url" content="<?php echo esc_attr( $url ); ?>" />
+		<?php
+	}
+	add_action( 'wp_head', 'inject_op_markup_meta_tag' );
```
everton-rosario: Same thing about "op" usage. Maybe we should drop this OP from everywhere.

facebook-instant-articles.php
```diff
@@ -342,20 +334,47 @@ function inject_url_claiming_meta_tag() {
...
+
 	// Initialize the Instant Articles meta box.
 	Instant_Articles_Meta_Box::init();
 
-	// Initialize the Instant Articles publisher.
-	Instant_Articles_Publisher::init();
+	function op_markup_version( ) {
+		$post = get_post();
+
+		if (isset($_GET['op']) && $_GET['op']) {
```
everton-rosario: 2 things:
- Formating nit, should be `if ( isset( $_GET[ 'op' ] ) && $_GET[ 'op' ] ) {`
- Same thing about "op" usage. Maybe we should drop this OP from everywhere.

facebook-instant-articles.php
```diff
@@ -342,20 +334,47 @@ function inject_url_claiming_meta_tag() {
...
+			// Transform the post to an Instant Article.
+			$adapter = new Instant_Articles_Post( $post );
+			$article = $adapter->to_instant_article();
+			echo $article->render(null, true);
```
everton-rosario: 💅  
`echo $article->render( null, true );`

facebook-instant-articles.php
```diff
@@ -342,20 +334,47 @@ function inject_url_claiming_meta_tag() {
...
+
+			die();
+		}
+	}
+	add_action( 'wp', 'op_markup_version' );
```
everton-rosario: Same thing about "op" usage. Maybe we should drop this OP from everywhere.

js/instant-articles-wizard.js
```diff
@@ -1,135 +0,0 @@
-function instant_articles_wizard_load ( data ) {
```
everton-rosario: Good code cleanup! 👍

tests/Facebook/InstantArticles/Transformer/WPTransformerTest.php
```diff
@@ -73,6 +73,7 @@ public function testTransformerLikeWPContent()
 
         $this->assertEquals($expected, $result);
         // there must be 3 warnings related to <img> inside <li> that is not supported by IA
+        // And 1 warning related to the getter
```
everton-rosario: Are you sure this comment should be added?
If there are 3 related to img + 1, the test assertion should be 4.
I guess this came from some dirty merge.

diegoquinteiro: Makes sense =) I'll remove it.

facebook-instant-articles.php
diegoquinteiro: Will replace "op" by "ia_markup"

meta-box/meta-box-template.php
```diff
@@ -18,135 +18,65 @@
...
-<?php if ( ! $should_submit_post ) : ?>
+<?php if ( $adapter->should_submit_post()  ) : ?>
+<p>
+	<b>
+		<span class="dashicons dashicons-yes"></span>
+		This post will be available as an Instant Article.
+	</b>
+</p>
+<hr>
+<?php elseif ( ! $should_submit_post ) : ?>
```
diegoquinteiro: Rename this variable

meta-box/meta-box-template.php
```diff
@@ -18,135 +18,65 @@
...
 <?php elseif ( ! $published ) : ?>
 <p>
 	<b>
 		<span class="dashicons dashicons-media-document"></span>
-		Your post will be submitted to Instant Articles once it is published.
+		This post will be submitted to Instant Articles once it is published.
+	</b>
+</p>
+<hr>
+<?php elseif ( ! $article->getHeader() || ! $article->getHeader()->getTitle() ) : ?>
```
diegoquinteiro: Move to a warning on the SDK

meta-box/meta-box-template.php
```diff
@@ -18,135 +18,65 @@
...
+		<span class="dashicons dashicons-no-alt"></span>
+		This post will not be submitted to Instant Articles because it is missing a title.
+	</b>
+</p>
+<hr>
+<?php elseif ( count( $article->getChildren() ) === 0 ) : ?>
+<p>
+	<b>
+		<span class="dashicons dashicons-no-alt"></span>
+		This post will not be submitted to Instant Articles because it is missings content.
```
diegoquinteiro: Typo (missings)

meta-box/meta-box-template.php
```diff
@@ -18,135 +18,65 @@
...
+<hr>
+<?php elseif ( ! $article->getHeader() || ! $article->getHeader()->getTitle() ) : ?>
+<p>
+	<b>
+		<span class="dashicons dashicons-no-alt"></span>
+		This post will not be submitted to Instant Articles because it is missing a title.
+	</b>
+</p>
+<hr>
+<?php elseif ( count( $article->getChildren() ) === 0 ) : ?>
```
diegoquinteiro: Move to a warning on the SDK

facebook-instant-articles.php
```diff
@@ -404,4 +404,44 @@ function ia_markup_version( ) {
...
+				$url_encoded = urlencode($adapter->get_canonical_url());
+				$client->send(
+					"https://graph.facebook.com/?id=$url_encoded&scrape=true",
+					'POST',
+					'',
+					array(),
+					60
+				);
+				foreach ( $old_slugs as $slug ) {
+					$clone_post = clone $post;
```
vkama: should this clonning really be inside the foreach? since only the post name is getting changed, maybe you can clone it once, outside, if at all.

facebook-instant-articles.php
```diff
@@ -404,4 +404,44 @@ function ia_markup_version( ) {
 	add_action( 'wp', 'ia_markup_version' );
 
 	Instant_Articles_Wizard::init();
+
+	function rescrape_article( $post_id, $post ) {
+		$adapter = new Instant_Articles_Post( $post );
+		$old_slugs = get_post_meta( $post_id, '_wp_old_slug' );
+		if ( $adapter->should_submit_post() ) {
+			try {
+				$client = Facebook\HttpClients\HttpClientsFactory::createHttpClient( null );
```
vkama: is this now the only time we use the graph-sdk? if so, maybe we should consider either using builtin functions for a simple post or actually copying the class and dropping it as a dep?

facebook-instant-articles.php
```diff
@@ -404,4 +404,44 @@ function ia_markup_version( ) {
 	add_action( 'wp', 'ia_markup_version' );
 
 	Instant_Articles_Wizard::init();
+
+	function rescrape_article( $post_id, $post ) {
```
vkama: is the $post_id required? isnt that the same as $post->ID?

diegoquinteiro: It is not required, but these are the parameters passed to the hook by WordPress.

facebook-instant-articles.php
diegoquinteiro: good call, I'll create a PR for improving the performance here

PR #646, opened 2017-03-20
Fix "PHP Notice:  Undefined index: publish_with_warnings"

abdusfauzi:
This PR:

Fixes #
Fix "PHP Notice:  Undefined index: publish_with_warnings" on PHP7.1. Added verification if array_key_exists to check if the option is enabled or disabled.


diegoquinteiro:
This code is not valid in v4 anymore, as we drastically changed this method.

PR #656, opened 2017-03-27
Load more managed pages to choose from

### LINKED ISSUE #492: Page selection only returns 25 results without pagination options
# Expected Result
- Pagination for users pages or list all of them rather than 25
# Actual Result
- Lists 25 max pages without pagination option

---

*\* Version Info **
- Plugin version: 3.1.3
- WordPress version: 4.6.1
- PHP version: 5.6

ExileStudios:
This change allows loading up to 250 pages, rather than being restricted to only 25 pages to select from.

I manage over 25 pages and couldn't select a page I needed to setup Instant Articles with, so this alteration was necessary for my scenario. I figured I would contribute it for others who may have the same issue.

closes #492 

Ideally pagination would be integrated, but this resolves the problem until pagination can be created.


timjacobi:
Thank you for your contribution. We have recently rewritten this settings for this plugin and are not loading the available pages anymore.

PR #699, opened 2017-06-24
Improvements to "Submit for Review" section during plugin installation

### LINKED ISSUE #698: "Submit for Review" pane can block if post has no title

1. Have a self-hosted Wordpress blog, with drafts that do not contain a title
2. Go through the process of installing the Wordpress plugin
3. Reach the "Submit for Review" step

# Expected Result

* Expect to clearly understand what the status of those 5 articles that were automatically converted is
* Don't expect to see drafts or private blog posts in this step as it is unlikely that those are ready to submit for review or even meant to be published yet (or at all) to IA

# Actual Result

* May see less than 5 articles (if some are valid)
* Drafts or posts with no title don't show up at all (can't click on them)

image

---

**Version Info**

* Plugin version: 3.3.5
* WordPress version: 4.8
* PHP version: 5.5.30

Nickology:
This PR:

* [x] Shows the status of all 5 articles being automatically converted for review
* [x] Only take into account posts that have status 'publish'
* [x] Prevents posts with no title from not being editable  

Follows #

Relates to #

Fixes #698 

image

### DIFF
--- a/wizard/class-instant-articles-wizard-review-submission.php
+++ b/wizard/class-instant-articles-wizard-review-submission.php
@@ -33,7 +33,8 @@ class Instant_Articles_Wizard_Review_Submission {
 		return wp_get_recent_posts(
 		 	array(
 				'numberposts' => min( self::MIN_ARTICLES, 100 ),
-				'post_type' => $post_types
+				'post_type' => $post_types,
+				'post_status' => 'publish',
 			),
 			'OBJECT'
 		);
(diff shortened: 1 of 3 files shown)


timjacobi:
@Nickology Is this PR still valid? If so could you merge master into your branch please and test your changes?
Thanks

Nickology:
Thanks for the ping @timjacobi !
@everton-rosario mentioned that this flow was removed completely in v4 of the plugin. Closing this PR unless that's no longer the case.

PR #708, opened 2017-06-29
Do not process non-post pages. Fixes #707

### LINKED ISSUE #707: Trying to get property of non-object in fb-instant-articles/class-instant-articles-post.php
In version 4.0.1, I'm getting lots of the following PHP Notices in our debug log. Both of these notices always fire together.

Trying to get property of non-object in /wp-content/plugins/fb-instant-articles/class-instant-articles-post.php on line 79
Trying to get property of non-object in /wp-content/plugins/fb-instant-articles/class-instant-articles-post.php on line 892

I'm uncertain of exactly the code path that is leading to this error, though it is definitely not tied to the publication or pushing of articles to Facebook. We are getting it many times per minute from normal pageload activity.

* Plugin version: 4.0.1
* WordPress version: 4.8
* PHP version: 7.1.5

diegoquinteiro:
This PR prevents the plugin from trying to inject the meta-tag on non-post pages, which was causing NOTICES to be shown on the debug log.

Fixes #707

### DIFF
--- a/facebook-instant-articles.php
+++ b/facebook-instant-articles.php
@@ -366,7 +366,7 @@ if ( version_compare( PHP_VERSION, '5.4', '<' ) ) {
 		$post = get_post();
 
 		// If there's no current post, return
-		if (! $post) {
+		if ( ! $post ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


facebook-instant-articles.php
```diff
@@ -364,6 +364,12 @@ function inject_url_claiming_meta_tag() {
 	 */
 	function inject_ia_markup_meta_tag() {
 		$post = get_post();
+
+		// If there's no current post, return
+		if (! $post) {
```
everton-rosario: `if ( ! $post ) {`
💅

PR #709, opened 2017-06-29
Add cache layer for avoiding transforming the article at page render

diegoquinteiro:
This PR address the 10% performance regression reported on the WordPress plugin support forum. 

I'm adding a cache layer using a simple post meta to avoid transforming the article to Instant Article at every page render for calculating whether it's a valid IA (which we use to decide whether to include the ia-markup meta tag or not).

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -876,12 +876,12 @@ class Instant_Articles_Post {
 	 */
 	public function should_submit_post() {
 
-		$post = get_post( $this->get_the_id() );
-
 		$fb_page_settings = Instant_Articles_Option_FB_Page::get_option_decoded();
 		if ( isset( $fb_page_settings[ 'page_id' ] ) && !$fb_page_settings[ 'page_id' ] ) {
 			return false;
 		}
+
+		$post = $this->_post;
 
 		// Don't process if this is just a revision or an autosave.
 		if ( wp_is_post_revision( $post ) || wp_is_post_autosave( $post ) ) {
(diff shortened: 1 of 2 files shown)


class-instant-articles-post.php
```diff
@@ -899,20 +899,29 @@ public function should_submit_post() {
...
+		if ( post_password_required( $post ) ) {
+			return false;
+		}
+
+		// Allow to disable post submit via filter
+		if ( false === apply_filters( 'instant_articles_should_submit_post', true, $this ) ) {
+			return false;
+		}
+
+		$cache = get_post_meta( $this->get_the_id(), 'should_submit_post_content', true );
```
everton-rosario: Is this post-meta documented?

diegoquinteiro: This post meta is only used internally by the plugin, it is not on the public interface, so I'll add in-code documentation.

There are no filters being added by this PR.

diegoquinteiro:
Good point. I also need to invalidate the cache if someone creates custom rules (as it can make invalid transformations become valid).

Will work on this and update the PR.

vkama:
Great job on separating the warnings part!

🐎 🐎 🐎

diegoquinteiro:
Local benchmark results (Apache Benchmark, 100 requests, 10 threads):

- Avg request time on 4.0.1: **482.448 ms**
- Avg request time on 3.3.5: **384.420 ms**
- Avg request time with PR: **363.885 ms**

Issue #569, opened 2016-12-23
Transformer rules for Themify Video embed and image slider


leccestar:
Here are transformer rules for making video embed and image slider with captions work in Themify builder posts:

    { "rules": 
        [
            { "class": "PassThroughRule", "selector": "div"}, 
            { "class": "InteractiveRule", "selector": "div.video-wrap","properties" : 
                { "interactive.iframe" : { "type" : "string", "selector" : "iframe", "attribute": "src"}}},
            { "class": "SlideshowRule", "selector": "ul.themify_builder_slider"},
            { "class": "SlideshowImageRule", "selector": "li", "properties" : 
                { 
                    "image.url" : 
                    { "type" : "string", "selector" : "img", "attribute": "src"
                    },
                    "caption.title" : 
                    { "type" : "string", "selector" : "img", "attribute": "alt"
                    }
                }
            }
        ]
    }

DJUK16:
Will this work with the slider pro plugin or is it just the main Themify builder plugin?

leccestar:
@DJUK16 it is for standard slider and video in the Themify Builder plugin.

Also I updated the code a bit:

{ "rules": 
        [
            { "class": "PassThroughRule", "selector": "div"}, 
            { "class": "InteractiveRule", "selector": "div.video-wrap","properties" : 
                {   "interactive.url" : { "type" : "string", "selector" : "iframe", "attribute": "src"},
                    "interactive.height" : { "type" : "int", "selector" : "iframe", "attribute": "height"},
                    "interactive.width"  : { "type" : "int", "selector" : "iframe", "attribute": "width"}
                }
            },
            { "class": "SlideshowRule", "selector": "ul.themify_builder_slider"},
            { "class": "SlideshowImageRule", "selector": "li", "properties" : 
                { 
                    "image.url" : 
                    { "type" : "string", "selector" : "img", "attribute": "src"
                    },
                    "caption.title" : 
                    { "type" : "string", "selector" : "img", "attribute": "alt"
                    }
                }
            }
        ]
    }

DJUK16:
Did this work successfully for you just im having some issues finding a good way to display images that looks nice both website side and FBIA side.

leccestar:
Yes I'm happy with how it looks on my blog (Themify theme and builder) and how it then looks on FBIA.

DJUK16:
[...]

PR #772, opened 2017-10-02
Add support for themify

timjacobi:
Add support for Themify plugin.

See #569 for details.

### DIFF
--- a/compat/themify-rules-configuration.json
+++ b/compat/themify-rules-configuration.json
@@ -1,0 +1,33 @@
+{
+  "rules": [
+    { "class": "PassThroughRule", "selector": "div" },
+    {
+      "class": "InteractiveRule",
+      "selector": "div.video-wrap",
+      "properties": {
+        "interactive.iframe": {
+          "type": "string",
+          "selector": "iframe",
+          "attribute": "src"
+        }
+      }
+    },
+    { "class": "SlideshowRule", "selector": "ul.themify_builder_slider" },
+    {
+      "class": "SlideshowImageRule",
+      "selector": "li",
+      "properties": {
+        "image.url": {
+          "type": "string",
+          "selector": "img",
+          "attribute": "src"
+        },
+        "caption.title": {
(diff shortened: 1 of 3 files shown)


compat/themify-rules-configuration.json
```diff
@@ -0,0 +1,33 @@
+{
+  "rules": [
+    { "class": "PassThroughRule", "selector": "div" },
+    {
+      "class": "InteractiveRule",
+      "selector": "div.video-wrap",
+      "properties": {
+        "interactive.iframe": {
```
pestevez: I think the correct property is `interactive.url`, as with `interactive.iframe` the markup I get is:
```html
<iframe class="no-margin">https://www.youtube.com/embed/waM20ewLj34?feature=oembed</iframe>
```
However, if I only change the property to `interactive.url`, the resulting markup does NOT contain the iframe element (not sure if this is an existing bug or the result of missing required properties.
The following properties worked well for me:
```javascript
"properties": {
        "interactive.url": {
          "type": "string",
          "selector": "iframe",
          "attribute": "src"
        },
        "interactive.height": {
          "type": "int",
          "selector": "iframe",
          "attribute": "height"
        },
        "interactive.width": {
          "type": "int",
          "selector": "iframe",
          "attribute": "width"
        }
      }
```

compat/themify-rules-configuration.json
```diff
@@ -0,0 +1,33 @@
+{
+  "rules": [
+    { "class": "PassThroughRule", "selector": "div" },
```
pestevez: If the video has the `Caption` property specified I still get a warning. I used the following rule to get rid of it:
```javascript
{ "class": "IgnoreRule", "selector": "div.video-caption" }
```

pestevez: I will add a PR to the SDK repo to support captions for Interactive elements since the IA format supports it.

iuzairjan:
fix this old shitty plugin @facebook

PR #775, opened 2017-10-03
Adds caching to instant_articles_embed_oembed_html

### LINKED ISSUE #685: Global wordpress performances and oEmbeds
I was investigating a performance issues on my websites (two medium audiences), and it seems that the plugin waste of lot of ressources and uselessly break the wordpress oembed discover cache in a way I don't yet understand.

## The problem

We use a log of self embeds (using a block to promote our own article on another article) and our /feed is not in the cache (30 posts).

So for each user asking for /feed and for each post containing a embed, it generate one "local" request in order to discover the oembed iframe URL. A lot of "internal" http requests for just one client.

Wordpress seems to usually have a cache for oembed, but the code in facebook-instant-articles/embeds.php to not use that cache for my /feed.

## First investigation / The place where it happens?

The function `instant_articles_embed_oembed_html` make use of  `wp_oembed->get_provider` which use `discover` which make external request to call the embeded URL in order to find the embed iframe URL. A parameter exists to force non-discovery, but is not used.

The embeds.php file is actually included on all the request, `remove_all_filters( 'embed_oembed_html' );` is then run on each requests and not only instant articles endpoint, potentially breaking other things.

Same for `add_filter( 'embed_oembed_html', 'instant_articles_embed_oembed_html', 10, 4 );`

---

**Version Info**

* Plugin version: 3.3.3
* WordPress version: 4.7.5
* PHP version: PHP 5.6.30-0+deb8u1

emrikol:
For each oEmbed on a page, an uncached remote request is made to check the oEmbed provider's URL.  This patch adds a level of caching to stop from having to check each oEmbed on every page load.

It's unlikely that an oEmbed provider will change for a specific URL, so there shouldn't be any need to specify an expiration on the good cached data.

This PR:

* [x] Adds caching to instant_articles_embed_oembed_html

Fixes #685

### DIFF
--- a/embeds.php
+++ b/embeds.php
@@ -24,30 +24,47 @@ remove_all_filters( 'embed_oembed_html' );
  */
 function instant_articles_embed_oembed_html( $html, $url, $attr, $post_id ) {
 
-	if ( ! class_exists( 'WP_oEmbed' ) ) {
-		include_once( ABSPATH . WPINC . '/class-oembed.php' );
+	$cache_key = md5( $url . ':instant_articles_oembed_provider' );
+	$provider_name = get_transient( $cache_key );
+
+	if ( false === $provider_name ) {
+		if ( ! class_exists( 'WP_oEmbed' ) ) {
+			include_once( ABSPATH . WPINC . '/class-oembed.php' );
+		}
+
+		// Instead of checking all possible URL variants, use the provider list from WP_oEmbed.
+		$wp_oembed = new WP_oEmbed();
+		$provider_url = $wp_oembed->get_provider( $url );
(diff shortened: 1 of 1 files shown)


emrikol:
Looks related to #337 too

timjacobi:
LGTM, thanks for your contribution!

sboisvert:
Hi @timjacobi I was just wondering when the next version (that included this fix) was planned for (to see if I should patch our current versions or wait for a new version) Thanks!

timjacobi:
We currently don’t have a release date so I recommend to patch for now.

Issue #791, opened 2017-10-30
Articles doesn't publish to Instant Articles (Publishing Tools)


rothanachoun:
# Steps required to reproduce the problem

1. Create a new post
2. After success we got
. This post will be available as an Instant Article.
. This post was transformed into an Instant Article with no warnings [Toggle debug information]

# Expected Result

* Published into Instant Articles (Publishing Tools)

# Actual Result

* Didn't see in Instant Articles (Publishing Tools)

---

**Version Info**

* Plugin version: 4.0.5
* WordPress version: 4.8.2
* PHP version: 5.5.9

timjacobi:
Thank you for reporting this. I'm just looked into this and can reproduce it. Will post more info once I've got to the bottom of it.

timjacobi:
Currently direct import is not supported. If you share the article on your page it will be scraped and made available to you in the _Publishing Tools_.

rothanachoun:
I got it. Thanks for you help.

timjacobi:
If there is anything else let's continue in #792

PR #792, opened 2017-10-30
Change meta box and readme so user doesn't think IAs are actually submitted to Facebook

timjacobi:
Due to the [recent change](https://developers.facebook.com/blog/post/2017/07/18/graph-api-v2.10) in the Graph API articles are not being published to Facebook automatically anymore but rather have to be shared first. See also #791 

This updates all messaging to account for this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,9 @@ Facebook requires a minimum number of articles in your feed before they will rev
 More likely than not, this is because there is markup in the body of your post that is not mapped to a recognized Instant Article component. On the “Edit Post” screen for your post, look for additional information about the *transformed* output shown within the **Facebook Instant Articles** module located at the bottom of the screen.
 
 **Why doesn't my post appear in the list of Instant Articles in the publisher tools?**
-In order for your post to be imported you have to share it on your page first.
-
+Your posts are imported to your library when they are shared on Facebook for the first time.
+ 
+Alternatively, you can trigger a manual scrape by pasting your URL on our [Share Debugger](http://developers.facebook.com/tools/debug)
 **In the Instant Articles module for my post, what does the “This post was transformed into an Instant Article with some warnings” message mean?**
 
 When transforming your post into an Instant Article, this plugin will show warnings when it encounters content which might not be valid when published to Facebook. When you see this message, it is recommended to resolve each warning individually.
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ You can propose popular transformer rules to be included in the plugin by [sugge
 
 **How do I post articles to Instant Articles after plugin is installed?**
 
-You can re-publish existing articles (simply edit + save) or post new articles in order to make them available as Instant Articles. After you have 5 articles added and shared on Facebook, you will be able to submit them for review.
+In order to import your posts to your Instant Articles library on Facebook you need to either share them on your page or use the [Sharing Debugger](https://developers.intern.facebook.com/tools/debug/sharing/) to scrape them. After you have 10 articles imported, you will be able to submit them for review.
 
 **How do I change the feed slug/URL if I'm using the RSS integration?**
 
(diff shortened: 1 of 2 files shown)


meta-box/meta-box-template.php
```diff
@@ -22,39 +22,39 @@
 <p>
 	<b>
 		<span class="dashicons dashicons-yes"></span>
-		This post will be available as an Instant Article.
+		This post will be available as Instant Article.
```
everton-rosario: + once shared on Facebook.

README.md
```diff
@@ -57,7 +60,7 @@ You can propose popular transformer rules to be included in the plugin by [sugge
 
 **How do I post articles to Instant Articles after plugin is installed?**
 
-You can re-publish existing articles (simply edit + save) or post new articles in order to submit them to Instant Articles. After you have 10 articles added, you will be able to submit them for review.
+You can re-publish existing articles (simply edit + save) or post new articles in order to make them available as Instant Articles. After you have 5 articles added and shared, you will be able to submit them for review.
```
everton-rosario: ... After you have 5 articles added and shared on Facebook, you will...

timjacobi:
@everton-rosario this PR is only to update the messaging. I agree we should remove the API calls but I will do that in a separate PR.

README.md
diegoquinteiro: I think now it's 10 indeed: https://developers.facebook.com/docs/instant-articles/quickstart#review

In fact, now you don't need to edit+save your posts to submit - it doesn't work anymore. We should let them know that new articles will be submitted when shared and old articles can be re-scraped through the Share Debugger.

meta-box/meta-box-template.php
diegoquinteiro: Maybe:

"This post will be available as Instant Article when shared on Facebook"

meta-box/meta-box-template.php
```diff
@@ -22,39 +22,39 @@
...
+		This post will not be available as Instant Article due to a rule created in your site.
 	</b>
 </p>
 <hr>
 <?php elseif ( ! $published ) : ?>
 <p>
 	<b>
 		<span class="dashicons dashicons-media-document"></span>
-		This post will be submitted to Instant Articles once it is published.
+		This post will be available as Instant Article once it is published.
```
diegoquinteiro: Here too:

"This post will be available as Instant Article once it is published and shared on Facebook"

README.md
```diff
@@ -39,6 +39,9 @@ Facebook requires a minimum number of articles in your feed before they will rev
 
 More likely than not, this is because there is markup in the body of your post that is not mapped to a recognized Instant Article component. On the “Edit Post” screen for your post, look for additional information about the *transformed* output shown within the **Facebook Instant Articles** module located at the bottom of the screen.
 
+**Why doesn't my post appear in the list of Instant Articles in the publisher tools?**
+In order for your post to be imported you have to share it on your page first.
```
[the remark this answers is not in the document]
timjacobi: I like it!

README.md
```diff
@@ -57,7 +61,7 @@ You can propose popular transformer rules to be included in the plugin by [sugge
 
 **How do I post articles to Instant Articles after plugin is installed?**
 
-You can re-publish existing articles (simply edit + save) or post new articles in order to submit them to Instant Articles. After you have 10 articles added, you will be able to submit them for review.
+You can re-publish existing articles (simply edit + save) or post new articles in order to make them available as Instant Articles. After you have 5 articles added and shared on Facebook, you will be able to submit them for review.
```
diegoquinteiro: The first sentence is still inaccurate. Editing + saving the article does nothing anymore. You need to share it (even without editing+saving). Check my previous comment. Also, we now require 10 articles, not 5.

meta-box/meta-box-template.php
```diff
@@ -22,39 +22,39 @@
 <p>
 	<b>
 		<span class="dashicons dashicons-yes"></span>
-		This post will be available as an Instant Article.
+		This post will be available as Instant Article once it is published and shared on Facebook.
```
diegoquinteiro: At this condition the article is already published, so maybe:

```
This post will be available as Instant Article once it is shared on Facebook.
```

PR #797, opened 2017-10-31
Display error message if meta box can't be loaded

### LINKED ISSUE #770: Plugin loading forever when configuration not working

1. Define custom rules below
2. Create a new article
```json
{
  "rules":[
    {
      "class": "MyRule",
      "selector": "strong img",
      "properties" : {
        "image.url" : {
          "type" : "string",
          "selector" : "img",
          "attribute": "src"
        }
      }
    }
  ]
}
```

# Expected Result

* Generic or specific error message is displayed

# Actual Result

* Loading spinner displays forever and no error is displayed

You can observe an error 500 in the network tab.

---

**Version Info**

* Plugin version: 4.0.5
* WordPress version: 4.8.2
* PHP version: 5.6.31

timjacobi:
This PR:
See #770 for initial problem. This PR adds a generic error message that is displayed if the meta box can't be loaded.

fixes #770

### DIFF
--- a/meta-box/meta-box-error.php
+++ b/meta-box/meta-box-error.php
@@ -7,4 +7,4 @@
  * @package default
  */
 ?>
-<p>An unkown error occurred while transforming the article. Please verify your rules confgiuration is correct.</p>
+<p>An unkown error occurred while transforming the article. Please verify your rules configuration is correct.</p>
--- a/meta-box/meta-box-error.php
+++ b/meta-box/meta-box-error.php
@@ -1,0 +1,10 @@
+<?php
+/**
+ * Facebook Instant Articles for WP.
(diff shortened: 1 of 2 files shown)


meta-box/meta-box-error.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * Facebook Instant Articles for WP.
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @package default
+ */
+?>
+<p>An unkown error occurred while transforming the article. Please verify your rules confgiuration is correct.</p>
```
diegoquinteiro: typo on "confgiuration" 💅

Issue #714, opened 2017-07-02
404 page not found when i share new articles (after update)


tavyDesign:
# Steps required to reproduce the problem

1. when i try to share new posts, i get 404 eror from website, in facebook cache
2. all of this is happend after last update and insert that ia_markup
3. 

# Expected Result

* 

# Actual Result

* 

---

**Version Info**

* Plugin version: 4.0.2
* WordPress version: 4.8
* PHP version: 7

diegoquinteiro:
Hi @tavyDesign,

Thanks for reporting. The change that might be causing your issue is that we now trigger a scrape of your article when you save it. If you have a cache or async replication in place in a way that your post is still unavailable to Facebook's scrapper by that time the action `save_post` happens on WordPress, it might scrape a 404 instead.

We'll evaluate the best way to address this use case. Meanwhile you can always manually re-trigger the scrape of the article again after a couple of seconds by using the Share Debugger UI.

tavyDesign:
Is not possible to make it on post publish? After link is available?

PR #798, opened 2017-11-01
remove article rescrape code

timjacobi:
This PR:
Removes the code that triggers a rescrape for an article. Due to a recent change in the Graph API this code does not work anymore.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,6 @@
   "require":
   {
     "php": ">=5.4",
-    "facebook/graph-sdk": "~5.0",
     "apache/log4php": "2.3.0",
     "facebook/facebook-instant-articles-sdk-php": "^1.6.0",
     "facebook/facebook-instant-articles-sdk-extensions-in-php": "^0.1.0",
--- a/facebook-instant-articles.php
+++ b/facebook-instant-articles.php
@@ -419,44 +419,6 @@ if ( version_compare( PHP_VERSION, '5.4', '<' ) ) {
 
 	Instant_Articles_Wizard::init();
 
-	function rescrape_article( $post_id, $post ) {
-		$adapter = new Instant_Articles_Post( $post );
-		$old_slugs = get_post_meta( $post_id, '_wp_old_slug' );
-		if ( $adapter->should_submit_post() ) {
-			try {
-				$client = Facebook\HttpClients\HttpClientsFactory::createHttpClient( null );
-				$url_encoded = urlencode($adapter->get_canonical_url());
-				$client->send(
-					"https://graph.facebook.com/?id=$url_encoded&scrape=true",
-					'POST',
-					'',
-					array(),
-					60
-				);
-				foreach ( $old_slugs as $slug ) {
-					$clone_post = clone $post;
-					$clone_post->post_name = $slug;
-					$clone_adapter = new Instant_Articles_Post( $clone_post );
-
-					$url_encoded = urlencode($clone_adapter->get_canonical_url());
-					$client->send(
-						"https://graph.facebook.com/?id=$url_encoded&scrape=true",
-						'POST',
-						'',
-						array(),
-						60
-					);
-				}
-			} catch ( Exception $e ) {
-				Logger::getLogger( 'instantarticles-wp-plugin' )->error(
-					'Unable to submit article.',
-					$e->getTraceAsString()
-				);
-			}
-		}
-	}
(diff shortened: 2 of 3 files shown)


pestevez:
Will the removal of the graph SDK dependency affect the use of `Facebook\Exceptions\FacebookResponseException` in file class-instant-articles-meta-box.php?

timjacobi:
Yes. That `use` statement is obsolete now :)

pestevez:
Fixes #714

DrLightman:
@timjacobi 
>Removes the code that triggers a rescrape for an article. Due to a recent change in the Graph API this code does not work anymore.

Thanks for the heads up. Can I ask you how did you learn about that? I've a site WP site with IA plugin and I was notified of some misbehaviour with posts not appearing in instant library and this could have shed some light on the issue.

timjacobi:
There is a post with release notes for the Graph API somewhere where they say that they deprecated it.

jerclarke:
Hey guys, this is the **DEFINITELY WRONG** way for this change to be communicated and explained. 

For starters, this change, along with a detailed explanation of the Graph API changes, should have been on  [the Instant Articles Blog](https://developers.facebook.com/ia/blog) becuase it is enormous and important and that blog promises to keep us up to date. 

I can imagine why you wouldn't want that though, because this change is **uniquely terrible for users and publishers, and deeply embarrassing to Facebook**. Still though, at least here on Github, it needed to be explained more usefully.

You should definitely be able to link to the place where the Graph API change was announced. 

You should also be able to explain how the change made it impossible for you to replace this feature with the new version of the API, since it cripples the plugin to have this feature removed. 

You should also have stated all the different effects this change would have had, for example that unlike before, where Instant Articles would update when the post was updated, now they will remain in the cache, stale and wrong, for 30 days. For more details see #878 

This change creates fake news by making updates to articles not show on Facebook. If we correct an article, our users will see the wrong version. By not making this clear to us as publishers, you left us ignorant of the problem, and guilty of false news without even knowing it.

jerclarke:
So, now I have looked through all the Graph API changes in the changelog, going back to 2016, and none of them imply that this should be broken, or that this feature should have been removed from this plugin. 

* [v. 2.1.2 - Released January 30, 2018](https://developers.facebook.com/docs/graph-api/changelog/version2.12): No mention of scraping
* [v. 2.11 - Released November 7, 2017](https://developers.facebook.com/docs/graph-api/changelog/version2.11): No mention of scraping
* [v. 2.10 - Released July 18, 2017](https://developers.facebook.com/docs/graph-api/changelog/version2.10): **Does mention scraping!** Described below
* *previous changelogs*: No mention of scraping

So it looks like [v. 2.10 - Released July 18, 2017](https://developers.facebook.com/docs/graph-api/changelog/version2.10) is the only sensible announcement that @timjacobi could have been referring to. 

**Here is the section that relates to scraping**

> URLs
> `POST /{url}` - This node now requires a valid access token.
> 
> `GET /{url}` - `og_object` is no longer returned by default but must be included in the fields parameter.
>
>` /{url}` Scraping - `GET` requests will no longer scrape the URL. `POST /{url}` must now be used to scrape the URL.

Now, this was all in the "90-Day Breaking Changes" section, which, because the release date was July 18 2017, means that this change to the `/{url}` node would have taken effect on October 16, 2017. This would then, logically, be the date when the code in this plugin stopped working, and about a month later this "fix" was applied to remove the code which no longer functioned. That all makes sense as a timeline, though it's still disturbing that the developers of this plugin didn't pre-empt the problem, having been given a 90 day warning, and working for the same company. Still, I won't hound you for that, I'm looking for a fix, not to blame people for doing their best.

**But how does that API change mean that we can't have scraping in this plugin?!?**

I can see that it broke, but why can't you fix it?

Based on reading the code, it doesn't seem like the GET/POST change is the culprit, as the code deleted from the plugin was already using POST. This leads me to believe that what broke the plugin was **"This node now requires a valid access token."**

So I don't even know exactly what that implies or entails, but I know that now I want to have an access token, and I want to use it to keep my posts up to date automatically the way this plugin kept them up to date before. 

**What is involved in getting an access token, and why can't this plugin accomplish it for us?**

Was this problem discussed in a meeting that we don't have access to? Is it impossible?

It seems like this is the specific quesetion that should have been discussed in this ticket, or at least there should have been an explanation of it, along with a link to the changelog. 

Looks to me like the feature is still there, and this is something that can be fixed.

PR #831, opened 2017-12-05
Fix the way filters are applied in `get_cover_media`

rebeccahum:
Hiya!

It would be more ideal to allow for the default behavior to execute and then, users can add the filter to overwrite `$cover_media` if they wish to do so.

props @sboisvert

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -517,6 +517,7 @@ class Instant_Articles_Post {
          * @since 0.1
          * @param Image     $cover_media  The cover media object.
          * @param int       $post_id      The current post ID.
+         */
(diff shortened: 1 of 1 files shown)


vkama:
Hi @rebeccahum, what exactly your code fixes? Is this just a refactor or is there a bug?

rebeccahum:
@vkama It was a bug, as we discovered that one of our VIP clients had a fatal upon upgrading:

```
Uncaught Error: Call to a member function get_the_featured_image() on null
```

We figured out that it was due to the order of execution of events, as there is no way to know what the default value of is `$cover_media` prior to deciding if to filter it.

sboisvert:
I love getting updates on random things like this :)

PR #833, opened 2017-12-05
Adding UTF-8 support (or any encoding)

sufrjo:
adding an option to send encoding to the AMPArticle::create function to use it to create UTF-8 instant articles

This PR:

* [x] 
* [ ] 
* [ ] 

Follows #

Relates to #

Fixes #


vkama:
Sorry, this is very low quality code and with hardcoded languages.

PR #870, opened 2018-01-31
[bugfix] calling get_article_style to apply the style filter

### LINKED ISSUE #864: get_article_style() method not implemented in class-instant-articles-post.php
## Steps required to reproduce the problem

The method `get_article_style()` that is part of the class-instant-articles-post.php is not implemented (it is never called). I tried hooking into the `instant_articles_style` filter, but since the `get_article_style()` method is not used anywhere, I am not able to override the article style set in the plugin settings in the WP dashboard. 

## Expected Result
Hooking into the `instant_articles_style` filter should override the default settings for the `fb:article_style` meta tag that is set in the WP dashboard. 

## Actual Result 
Nothing is written to the error log when the code below is called. `WP_DEBUG` and `WP_DEBUG_LOG` are set to true. 

```php
function fc_fbia_style( $template, $post_obj ) {
    error_log( 'this is running' );
    //... rest of function
}
add_action( 'instant_articles_style', 'fc_fbia_style', 10, 2 );
```

## Version Info

* Plugin version: 4.0.6
* WordPress version: 4.9.2
* PHP version: 7.0.1

vkama:
This PR fixes #864 

It now correctly applies the filter style

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -991,9 +991,8 @@ class Instant_Articles_Post {
 			$this->instant_article->withStyle( 'default' );
 		}
 
-		$override_style = $this->get_article_style();
-		if ($override_style != 'default') {
-			$this->instant_article->withStyle($override_style);
+		if (has_filter('instant_articles_style')) {
+			$this->instant_article->withStyle($this->get_article_style());
 		}
 
 		if ( isset( $settings['copyright'] ) && ! empty( $settings['copyright'] ) ) {
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -991,6 +991,11 @@ class Instant_Articles_Post {
 			$this->instant_article->withStyle( 'default' );
 		}
 
+		$override_style = $this->get_article_style();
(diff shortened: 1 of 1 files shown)


class-instant-articles-post.php
```diff
@@ -991,6 +991,10 @@ public function set_appearance_from_settings() {
 			$this->instant_article->withStyle( 'default' );
 		}
 
+		if (has_filter('instant_articles_style')) {
```
everton-rosario: Its better to remove the method `get_article_style()` and call inline the `apply_filters`.

Also add a documentation block before this method call `apply_filters`
```
                /**
		 * Apply the default filter 'instant_articles_style' for the instant article.
		 *
		 * @since 4.0.6
                 * @param string                           $template                     Path to the current (default) template.
		 * @param Instant_Article_Post    $instant_article_post   The instant article post.
		 */
                 $article_style = apply_filters( 'instant_articles_style', 'default', $this );
                  if ( $article_style !== null && !Type::isTextEmpty( $article_style ) ) {
                      $instant_article->withStyle( $article_style );
                  }
```

PR #802, opened 2017-11-03
Adds hooks for ads and AN placement id

yeah-sir:
This PR:

* Adds hook for ads 
* Adds hook for Ads Network placement id

Follows #

Relates to #

Fixes #

PR #873, opened 2018-02-07
Implementing Ads Hooks

vkama:
The idea is to enable a developer to customize the way it shows ads. There are no hooks today.

If they want to change the AN placement ID based on some logic, they can't.
If they want to use AN for some article and Direct sold ads for another, they can't.

The code was started by @yeah-sir on #802 but since that PR started from his repo I can't update it.

Reviewers: I'd like you to be extra attentive when reviewing this, since this deals with monetisation. I tested this but you never know :)

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -737,12 +737,25 @@ class Instant_Articles_Post {
 			$height = intval( $dimensions_match[2] );
 		}
 
+		$source_of_ad = isset( $settings_ads['ad_source'] ) ? $settings_ads['ad_source'] : 'none';
+
+		$ad = $this->get_ads_for_ad_source($source_of_ad, $width, $height);
+
+		// Hook for Ads
+		if( has_filter('instant_articles_loading_ads') ){
+			$ad = apply_filters('instant_articles_loading_ads', $ad);
+		}
+		$header->addAd( $ad );
+
+		$this->instant_article->enableAutomaticAdPlacement();
+	}
+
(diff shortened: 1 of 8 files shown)


class-instant-articles-post.php
```diff
@@ -737,12 +737,27 @@ public function add_ads_from_settings() {
 			$height = intval( $dimensions_match[2] );
 		}
 
+		$source_of_ad = isset( $settings_ads['ad_source'] ) ? $settings_ads['ad_source'] : 'none';
+
+		$ad = $this->get_ads_for_ad_source($source_of_ad, $width, $height);
```
everton-rosario: What do you think of a oneliner with documentation:

```
    $ad = apply_filters( 
        'instant_articles_loading_ads', 
        $this->get_ads_for_ad_source( $source_of_ad, $width, $height)
    );
    $header->addAd( $ad );
```

Mind the code formatting of spaces ( param ).
Also add the doc block for this filter.

class-instant-articles-post.php
```diff
@@ -752,6 +767,11 @@ public function add_ads_from_settings() {
 				if ( ! empty( $settings_ads['fan_placement_id'] ) ) {
 					$placement_id = $settings_ads['fan_placement_id'];
 
+					// Hook for placement ID
```
everton-rosario: Same here, no need to the `if (has filter)`

```
	$placement_id = apply_filters( 'instant_articles_ads_placement_id', $placement_id);
```

Issue #895, opened 2018-03-07
Playbuzz Plugin new version using new embed code 


shaibo10:
## 

In the next version of Playbuzz WordPress Plugin, the plugin will use the new (v3.0.0) Playbuzz embed code, and you are not supporting it in your plugin

The new embed code will look like this:
```
<script>(function(d,s,id){var js,fjs=d.getElementsByTagName(s)[0];if(d.getElementById(id))return;js=d.createElement(s);js.id=id;js.src='https://embed.playbuzz.com/sdk.js';fjs.parentNode.insertBefore(js,fjs);}(document,'script','playbuzz-sdk'));</script>
<div class="playbuzz" data-id="fadf317e-edb0-4f1f-b9cb-d5a79a1a3541" data-show-info="false"></div>
```

Plus some attributes for supporting other WordPress plugins that use Playbuzz Embed v2

Testing it is easy:
1. Use Playbuzz Plugin version <= v1.0.6 
2. Copy this embed code to your wordpress post editor
3. Click on Facebook Instant Articles->Toggle debug information
4. Check if the the script in the iframe is "embed.playbuzz.com/sdk.js"
5. If not, you are not supporting the new Playbuzz Embed code v3.0

* Playbuzz Plugin version if installed:  <= v1.0.6 
* Plugin version: 4.0.6
* WordPress version: 4.9.4
* PHP version: 7.0.23

jerclarke:
There should really be a filter that external plugins can use to insert extra JSON into the transformer rules, that way this code could be in the Playbuzz plugin instead. 

Having to manage patches like this one is a senseless pain for the IA plugin devs. Setting up a good filter would save IA dev time as well as making it so WAY more plugins would add IA support on their end. 

We do this in our plugin by fetching the existing option value, pulling out the sub-array with the transformations, deconstructing them into a PHP array, adding our custom rules, then re-encoding them into JSON and replacing them into the sub-array. 

This is a pain and very fragile when done this way from our end. Would be much better if it was a filter inside **Instant Articles for Wordpress** that anyone could use.

everton-rosario:
Thats a great idea @jerclarke.
Would you be up for sending a PR with this approach?

jerclarke:
I'll consider it, but it would have to wait for a time when I really needed it. Since we already have our solution (albeit hacky) set up, I don't actually need this myself.

mburak:
Fixed by #896

PR #896, opened 2018-03-07
feat(playbuzz): support for the next major playbuzz plugin version

shaibo10:
This extra rule support Playbuzz plugin next version,
Playbuzz plugin next version support new Playbuzz embed code v3.0, for match better SEO and removes deprecated Playbuzz code dependencies

This PR: adding the ability to use the new Playbuzz plugin which use the new Playbuzz embed code, 
It won't break the support of the old Playbuzz embed code (v2).

Relates to #895


shaibo10:
@mburak can you look at it?

Issue #877, opened 2018-02-13
problem with lazy load


lubimow:
I'm using the Wp-rocket plugin. When I enabled lazy-load, images in the article instant articles began to appear in this form:   

` <img src="data:image/gif;base64,R0lGODdhAQABAPAAAP///wAAACwAAAAAAQABAEACAkQBADs="/>`

How can I fix this?

## Version Info

* Plugin version:  4.0.6 
* WordPress version: 4.9.4
* PHP version: 7.1

pestevez:
Hi @lubimow, thanks for reporting this issue. Since this is related to compatibility with a third-party plugin, our team will treat it as a feature request, so it will be prioritized with the rest of requests.

arunbasillal:
Hey guys, 

I am Arun Basil Lal, I work for WP Rocket and I just wanted to share couple of insights. 

WP Rocket already has checks in place to disable LazyLoading on feeds and doesn't show up on the default WordPress feeds. Here is the part of the code that corresponds to it: https://github.com/wp-media/wp-rocket/blob/master/inc/front/lazyload.php#L139

If you have any questions related to it, please do not hesitate to let us know. 

Best regards.

lubimow:
@pestevez Please read the comment above

maungthuu:
Hi @lubimow 
 
Put this code in Custom Transformer Rules. 

`{
    "rules": [{
        "class": "ImageRule",
        "selector": "img[data-src]",
        "properties": {
            "image.url": {
                "type": "string",
                "selector": "img",
                "attribute": "data-src"
            }
        }
    }]
}`

lubimow:
@maungthuu Thanks, this solved my problem

maungthuu:
@lubimow glad it helps.

Singletrackmark:
We have the same problem. I've added the custom rule above to our IA plugin. Do I have to go back and republish all affected articles now?

maungthuu:
@Singletrackmark you need to update affected articles again.

lubimow:
@maungthuu Hey. I again do not see a link to the image. Help, what to do?

PR #905, opened 2018-03-15
Add Image Rule For Lazy Loaded Images

maungthuu:
Fix For Lazy Load Image enabled websites.

See #877


maungthuu:
Hi @everton-rosario 

This is not a fix for WP Rocket Plugin. I believe it is common that Lazy-Loaded Images Feature replaces img tag with data-src attribute in the markup.

PR #909, opened 2018-03-26
Move admin scripts and styles so that they only show on fb instant articles page

andykillen:
This PR:

Moves the enqueue of the admin scripts and styles so that they only load on the admin page of instant articles.  

While the scripts & styles are very small, they are still connection requests that are not needed for every other admin page. By not loading them it will make the mobile loading time quicker, and also loading in any country where there is poor networking.

Follows #

Relates to #

Fixes #


GaryJones:
Needs updating to fix a merge conflict, but I think this is worthwhile.

Issue #925, opened 2018-04-19
Website Verification


Mohan2266:
I am getting this error while trying to submit my website URL 
"The provided URL is not being accepted because the site has minimal readership, which violates the Instant Article Policies (https://developers.facebook.com/docs/instant-articles/policy/). Please provide a different URL."

I install instant article plugin, kept page ID into it and everything done right. My blog is https.

everton-rosario:
Hello, @Mohan2266
Can you share the blog URL and FB Page ID?

domimvp:
Hi im having the same problem i have been using my site sinds last year and now im getting this error i dont know why :( any help will be much appreciated

cemmildon:
I'm having the same problem. Why we can't add our site?

basudevkunwar:
Hi im also having the same problem , my website is new, just created 3 days ago , and today trying to Connect my  Site on instant articles , there is i saw (**<b>The provided URL is not being accepted because the site has minimal readership, which violates the Instant Article Policies (https://developers.facebook.com/docs/instant-articles/policy/). Please provide a different URL.</b>) this warning.

micolore:
Hi im also having same problem. i think one page only add one domain . but  last month can add many website . when i remove already added website it's can't add again. why

nhavietnamese:
The provided URL is not being accepted because the site has minimal readership, which violates the Instant Article Policies (https://developers.facebook.com/docs/instant-articles/policy/). Please provide a different URL.
please help me solve this problem

Kumanova22:
The provided URL is not being accepted because the site has minimal readership, which violates the Instant Article Policies (https://developers.facebook.com/docs/instant-articles/policy/). Please provide a different URL.
i have this problem from switzerland hmmm? why , why , why???

mizansyl:
plz help  me why i getting message when i want verify fb
The provided URL is not being accepted because the site has minimal readership, which violates the Instant Article Policies (https://developers.facebook.com/docs/instant-articles/policy/). Please provide a different URL

mkasik:
Same problem. Any one can solve this?

taaariqul:
I also have the same problem.
Can any expert help with this?

jerclarke:
Okay team, I don't want to be a downer, but in case any of you think this is just an error we should of course take seriously the text of the message:

[...]

PR #931, opened 2018-04-30
Update Readme re:non-minimal readership

jerclarke:
Now that it is impossible to use this plugin with a site that has ["minimal readership" on Facebook](https://github.com/Automattic/facebook-instant-articles-wp/issues/925#issuecomment-385243511) it is vital that the plugin warns developers not to set it up unless their site already meets the guidelines. This PR is a first step to warn people that after all their hard work getting their site to validate as IA they will still likely be blocked because of minimal readership.

This PR:

* [x] 
* [ ] 
* [ ] 

Follows #

Relates to #925 

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Enable [Instant Articles for Facebook](https://developers.facebook.com/docs/inst
 
 This plugin adds support for Instant Articles for Facebook, which is a new way for publishers to distribute fast, interactive stories on Facebook. Instant Articles are preloaded in the Facebook mobile app so they load instantly.
 
+**Instant Articles for WP only works on sites with non-minimal readership. If you do not have an active Facebook page with  non-minimal readership your site will not be accepted to Instant Articles even if you configure this plugin correctly.**
+
 With the plugin active, a new menu will be available for you to connect to your Facebook Page and start publishing your Instant Articles. You'll also see the status of each Instant Articles submission on the edit page of your posts.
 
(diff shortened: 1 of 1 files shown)


jerclarke:
Any response to the obvious need that this addresses? Did you somehow explain this elsewhere?

GaryJones:
There was a merge conflict, but if you want to fix that up, then I'm happy to take another look :-)

jerclarke:
The PR wasn't really the point, as much as the abject failure of the plugin to warn people that it is useless if your FB page doesn't meet the minimum readership requirements for IA. 

Originally there was no minimum readership requirement, and all the docs, like this one, don't mention it at all:

https://developers.facebook.com/docs/instant-articles/wordpress-quickstart/

But this is disastrous, because for people with a non-qualifying page it can take many many hours of integration to get to the step where you will discover it was all wasted. 

The plugin page should express these limitations and how to check if you qualify BEFORE installing the plugin.

Last time I checked this was still a non-transparent process and there really was no way to check in advance if you qualify. This should be a major warning on the plugin description.

GaryJones:
Looking at https://github.com/Automattic/fb-instant-articles#access-to-instant-articles (this is the bit that has the merge conflict) then it seems that it's already been made more explicit what the requirements are for Instant Articles to be accessed?

jerclarke:
It's not clear. There's no indication of what will happen if your site is deemed to have minimal readership, and it's extremely important because at that point people will have wasted all of their effort leading up to that discovery. 

It's a core failure of IA, not part of this plugin, but there should be warnings in this plugin about it. 

IA is so dead that it hardly matters, no one new is actually trying to set it up. 

But I maintain that it's a specific failure of this plugin (and the linked documentation) that there are no warnings, since otherwise the docs for this plugin seem to tell you everything you need to know. 

If you want to know what i'm talking about, try to set up IA on a small website without a ton of traffic and you'll see how much work you do before you discover you can never use IA.

PR #937, opened 2018-05-13
Add token-less re-scrape with request signature

diegoquinteiro:
This PR aims to do trigger a token-less re-scrape of the URLs being updated to trigger a new Instant Article ingestion even if the user does not have a configured app ID/app secret/access token.

- [x] Create key generation/exposure logic
- [x] Gracefully degrade to require a token when OpenSSL lib is not available

### DIFF
--- a/class-instant-articles-signer.php
+++ b/class-instant-articles-signer.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * Facebook Instant Articles for WP.
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @package default
+ */
+
+/**
+ * Class responsible for handling the signature of graph rescrape calls
+ *
+ * @since 4.1
+ */
+class Instant_Articles_Signer {
+
+  const PUBLIC_KEY_OPTION = 'instant-articles-rescrape-public-key';
+  const PRIVATE_KEY_OPTION = 'instant-articles-rescrape-private-key';
+  const PUBLIC_KEY_PATH = '/.well-known/graph-api/apikey.pub';
+
+  public static function init() {
+    add_action( 'wp', array( 'Instant_Articles_Signer', 'output_public_key' ) );
+  }
+
+  public static function output_public_key() {
+    global $wp;
+    if ( $wp->request == self::PUBLIC_KEY_PATH ) {
+      echo self::get_public_key();
+      die();
+    }
+  }
+
+  public static function get_public_key() {
+    $public_key = get_option( self::PUBLIC_KEY_OPTION );
+    if ( ! $public_key ) {
+      self::gen_keys();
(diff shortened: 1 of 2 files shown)


diegoquinteiro:
Hi @jerclarke!

That's one of the changes we're testing to solve the problem, based on your feedback. But we're still rolling out the API side of it and we can only review/land this PR afterwards.

Would you like/be able to help testing it?

Thank you!

diegoquinteiro:
@everton-rosario This PR is now complete and ready for review.

facebook-instant-articles.php
```diff
@@ -563,4 +565,60 @@ function admin_notice__scrape_invalidation_success() {
 	}
 	add_action( 'post_updated', 'invalidate_scrape_on_update', 10, 3 );
 
+	function rescrape_article( $post_id, $post ) {
+		$adapter = new Instant_Articles_Post( $post );
+		$old_slugs = get_post_meta( $post_id, '_wp_old_slug' );
+		if ( $adapter->should_submit_post() ) {
+
```
everton-rosario: Remove extra empty line.

class-instant-articles-signer.php
```diff
@@ -0,0 +1,88 @@
...
+
+  public static function init() {
+    add_action( 'wp', array( 'Instant_Articles_Signer', 'output_public_key' ) );
+  }
+
+  public static function output_public_key() {
+    global $wp;
+    if ( $wp->request == self::PUBLIC_KEY_PATH ) {
+      status_header(200);
+      echo self::get_public_key();
```
everton-rosario: Shouldn't we add proper content-type?

everton-rosario: Check if we shouldn't allow more text into this response. So `the_content` should not have anything else but the key.

class-instant-articles-signer.php
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Class responsible for handling the signature of graph rescrape calls
+ *
+ * @since 4.1
+ */
+class Instant_Articles_Signer {
+
+  const PUBLIC_KEY_OPTION = 'instant-articles-rescrape-public-key';
+  const PRIVATE_KEY_OPTION = 'instant-articles-rescrape-private-key';
+  const PUBLIC_KEY_PATH = '.well-known/graph-api/apikey.pub';
```
everton-rosario: Is this a fixed place? Should we make it by default this path, but allow configuration in the settings page.

diegoquinteiro: This is a fixed place, yes. Will not be configurable.

class-instant-articles-signer.php
diegoquinteiro: We `die()` aftwards to ensure nothing else is coming in out the response.

facebook-instant-articles.php
diegoquinteiro: 💅

PR #941, opened 2018-05-20
Add support for GTM4WP

robbiet480:
This PR:

* [x] Adds support for [Google Tag Manager for WordPress by Duracelltomi (GTM4WP)](https://gtm4wp.com/)

### DIFF
--- a/compat.php
+++ b/compat.php
@@ -28,6 +28,13 @@ if ( ( defined( 'GAWP_VERSION' ) || function_exists( 'MonsterInsights' ) ) && ! 
 	$gawp->init();
 }
 
+// Load support for Google Tag Manager for WordPress by Duracelltomi.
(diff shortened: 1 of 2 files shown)


koconder:
@robbiet480 just subscribing to this PR, calling the plugin code directly might not be the best solution from gtm4wp, as any changes in the setup on the plugin might break things. Better to build support using hooks if possible.

everton-rosario:
I agree on adding a less coupled solution, but this is exactly the solution we have for compatibility with other plugins.
@robbiet480 also added a configuration style once the plugin is active or not, just like how we initially implemented for google analytics.

Just tested and it is good to go. 

Thanks, @robbiet480 


To cover the points you've mentioned @koconder , do you want to give it a try in a PR for refactoring this plugin dependency structure?

PR #952, opened 2018-06-28
Updating description for the plugin

Blakomen:
Making the requirements for using Instant Articles (a prerequisite for the WP plugin) a little clearer


jerclarke:
If you're going to have a bulleted list, why not put the minimal readership aspect into it? That's the most important part, and the aspect that will disqualify almost everyone.

ahr36 (approved):

Issue #953, opened 2018-06-29
ia:markup_url being output on home.php, author.php etc. because of get_post() instead of is_singular()


jerclarke:
## Steps required to reproduce the problem

1. Have the plugin set up and working
2. Visit homepage/author screen/term screen
3. View source and search for `ia:markup_url`

## Expected Result

Nothing, because these screens can't possibly be summarized as IA and they aren't.

## Actual Result 

Homepage header has this in it (POST_URL is the first post that was on my screen, but would vary depending on the arbitrary state of your loop globals):

```HTML
<meta property="ia:markup_url" content="https://{POST_URL}/?ia_markup=1" />
```

## What's happening? You're not fetching the post properly when testing if it's valid

This function is the culprit: `inject_ia_markup_meta_tag()`

It is used to insert the meta tag in the HTML of single posts so that it can be found by the Facebook scraper. It in turn calls `Instant_Articles_Post->should_submit_post()`, which seems to work properly and test the `$post` in various ways and return false if it's invalid (e.g. it's a page, it's unpublished etc.)

The problem, is that `inject_ia_markup_meta_tag()` doesn't do any testing at all to make sure it's on a single post screen (`is_singular()` or `is_single()`) and just passes the output of `get_post()` in, regardless of whether `get_post()` will return something that represents the current screen or not. 

In the case of `archive.php`,  `index.php` and friends, the output of `get_post()` during the header is totally arbitrary, and should never be relied upon. Any kind of loop may have been run already, in which case `get_post()` might return something random. Those screens have all kinds of posts, and obviously, none of them singularly represents the archive in a way that makes sense for IA to ever insert the `ia:markup_url` tag. 

Further, even on a `single.php` view (where `is_single()` returns true), using `get_post()` this way is foolish, as there may have been other loops running aside from the main post, thus populating `global $post` and impacting the output of `get_post()`. 

## Solution: Use is_singular() AND get_the_id()

AFAICT what you are really expecting here is this:

> get the post object for the single post that represents this view, then insert it's url in `ia:markup_url`

If this isn't a "single post view" then you don't want to even consider inserting the `ia:markup_url` tag, and if it **is** a "single post view", then the only post you are interested is the "official" one for this view, which is reliably returned by `get_the_id()`.

So this bit of code in `facebook-instant-articles.php`:

```PHP
function inject_ia_markup_meta_tag() {
    $post = get_post();

    // If there's no current post, return
    if ( ! $post ) {
        return;
    }
```

[...]

PR #954, opened 2018-06-29
Fix inject_ia_markup_meta_tag() to use is_singular() and get_queried_object_id()

jerclarke:
re: #953

Small fix to ensure that before inserting the `ia:markup_url` meta tag, we check that we're on a single-post screen (`is_singular()` which is true for posts, pages and CPT `single.php` views).  
Also uses with `get_post(get_queried_object_id())` rather than relying on globals (no argument or using `get_the_ID()`) so that we always have the right post.

This PR:

* [x] 
* [ ] 
* [ ] 

Follows #

Relates to #

Fixes #

### DIFF
--- a/facebook-instant-articles.php
+++ b/facebook-instant-articles.php
@@ -373,12 +373,10 @@ if ( version_compare( PHP_VERSION, '5.4', '<' ) ) {
 	 * @since 4.0
 	 */
 	function inject_ia_markup_meta_tag() {
-		$post = get_post();
-
-		// If there's no current post, return
-		if ( ! $post ) {
+		if ( !is_singular() )
 			return;
-		}
+		
+		$post = get_post( get_the_ID() );
 
 		// Transform the post to an Instant Article.
 		$adapter = new Instant_Articles_Post( $post );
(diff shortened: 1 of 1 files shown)


ahr36 (approved):

facebook-instant-articles.php
```diff
@@ -373,9 +373,11 @@ function inject_url_claiming_meta_tag() {
 	 * @since 4.0
 	 */
 	function inject_ia_markup_meta_tag() {
-		$post = get_post();
-
-		// If there's no current post, return
+		if ( !is_singular() )
+			return;
+		
+		$post = get_post( get_queried_object_id() );
```
GaryJones, line 379: Under what conditions would this now give a different result than it did before? Can you add an integration test to cover it?

facebook-instant-articles.php
```diff
@@ -373,9 +373,11 @@ function inject_url_claiming_meta_tag() {
 	 * @since 4.0
 	 */
 	function inject_ia_markup_meta_tag() {
-		$post = get_post();
-
-		// If there's no current post, return
+		if ( !is_singular() )
+			return;
```
GaryJones, line 377:
```suggestion
		if ( ! is_singular() ) {
			return;
		}
```

GaryJones (changes requested):

GaryJones:
Closing for lack of feedback.

PR #972, opened 2018-08-31
Fixes undefined 'display_warning_column' column notices for woocommer…

bampisSyk:
…ce order admin view

This PR:

Fixes the warnings produces by missing array key '$publishing_settings[ 'display_warning_column' ]'.
Key seems to be missing from woocommerce order admin page.


GaryJones:
If this is still an issue, it will need the merge conflict resolving.

PR #997, opened 2018-11-29
Removed fields from deprecated feature to enable comments/likes/share from media content

everton-rosario:
This adapts the Instant Articles Wordpress plugin to adapt to the deprecation of media feedback to individual media components of Instant Articles.

Commonly used attribute into the figure.
```
<figure data-feedback="fb:likes, fb:comments">
  <img src="http://mydomain.com/path/to/img.jpg" />
</figure>
```
This takes no effect and we stopped outputting in case this gets enabled. So the methods enabling/disabling data-feedback are set to do nothing now. 

On the plugin, we removed the setup page for this properties.

Relates to https://github.com/facebook/facebook-instant-articles-sdk-php/pull/347

### DIFF
--- a/class-instant-articles-post.php
+++ b/class-instant-articles-post.php
@@ -681,19 +681,6 @@ class Instant_Articles_Post {
 
 		$this->set_appearance_from_settings();
 
-		// This block sets as default likes and/or comments based on the configuration setup,
-		// and call $transformer->transform will consider the defaults before building the Elements
-		//
-		// Warning: if you are using pthreads or any other multithreaded engine, consider replicating this to all processes.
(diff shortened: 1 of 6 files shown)


VoiceOf1980:
But The Problem is Still Getting Error! when Updating Version 4.1.1 To version 4.2.0 . In My case can't render any articles From my Wordpress Dashboard To  instant articles pannale. All Articles get Warnings. 
After Rollback  Old Version 4.1.1 Its Working fine ! 
Also The Warning Is Still On Pluging Updates Page From Last Years 

> Warning: This plugin has not been tested with your current version of WordPress.

Issue #391, opened 2016-08-03
WPCOM Compat with plugins_loaded Action


douglas-johnson:
The function [instant_articles_load_compat](https://github.com/Automattic/facebook-instant-articles-wp/blob/master/facebook-instant-articles.php#L129) uses the action `plugins_loaded`

However on WordPress VIP, plugins are often loaded from the theme, especially if you want to use VIP's plugin versioning. That is how I have to load CoAuthors Plus, so the CAP compat layer is not working. 

May I suggest changing this action to use `after_setup_theme` or `init` since the theme has access to them? That solved this problem for me locally.

everton-rosario:
@douglas-johnson Thanks for bringing this scenario.
Can you send us a PR with this change and tests within VIP and on hosted WP?

Thanks for helping making this plugin better.

dashaluna:
Just hit this problem! God that was a pain. You'd need to load `wpcom-helper.php` file, where there is a solution to load functions in questions on `after_setup_theme`.

It would be great if there is a clear comment in the main plugin file while including `compat.php` ref: https://github.com/Automattic/facebook-instant-articles-wp/blob/master/facebook-instant-articles.php#L132 explaining that `wpcom-helper.php` is required for VIP, as it's not very straightforward.

PR #1001, opened 2018-12-20
Load functions which rely on `plugins_loaded` action hook in VIP Go environment

paulschreiber:
This PR:
Ensures plugins and localization are loaded in the VIP Go environment.

Relates to #391 #392.


paulschreiber:
@diegoquinteiro

PR #1007, opened 2019-01-06
Update Facebook SDK versions

paulschreiber:
Update Facebook Instant Articles SDK to 1.10.0 (from 1.9.0) and Facebook Instant Articles Extensions SDK to 0.2.1 (from 0.1.0).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,8 +25,8 @@
   "require":
   {
     "php": ">=5.4",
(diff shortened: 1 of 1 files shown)


paulschreiber:
@everton-rosario do you want to run `composer install` and update `composer.lock` as well?

everton-rosario:
Yes, we generally do it on package release.
I'm just running thru the rest of PR's merging in this cut so I can get it all in one.

PR #1011, opened 2019-01-10
Add option to disable IA on each post

### LINKED ISSUE #773: Add option to disable plugin for a single post
As requested in #567 it would be great if we could allow people disable publishing on a per article basis. A checkbox should be enough.

paulschreiber:
This PR:

* Adds an enabled/disabled toggle to the publish metabox
* Adds a filter to use this metadata to exclude posts from the IA feed
* Is adapted from the AMP-WP plugin's toggle (also GPLv2)

Fixes #773


paulschreiber:
@timjacobi

paulschreiber:
@diegoquinteiro

For (1), I'm not sure what you are requesting.

For (2), I'm guessing you want `rescrape_article()` called. Won't that happen already, since:
```
add_action( 'save_post', 'rescrape_article', 999, 2 );
```
…is set in [facebook-instant-articles.php:584](https://github.com/Automattic/facebook-instant-articles-wp/blob/master/facebook-instant-articles.php#L584)?

This will be faster if you fork off my branch and make the desired changes. I'm happy to give you write access to my fork if that's easier.

[Aside: `rescrape_article` is a pretty generic function name. Is this namespaced in a way I don't understand? If not, it would make more sense as a static method in some object.]

terrapop:
When will this be merged? This is essential. Two years and no change?

HRarsha:
> 

Hello @paulschreiber I've installed your plugin with composer installed and it's activated and setupped with page id, BUT I'm not seeing anything to enable or disable toggle for individual article? simply, it's almost same as the official, am I missing anything ? 

Anyway to see a screenshot of new options after activating the plugin? 

thanks in advance.

paulschreiber:
@HRarsha I did not create the plugin, not do I maintain it.

PR #1012, opened 2019-01-10
update composer modules

paulschreiber:
Updates composer.lock (with the results of `composer update`).


jeremeylduvall:
@everton-rosario I noticed you reviewed #1002. Would you be available to review this one as well?

PR #1013, opened 2019-01-11
Remove AMP support from Instant Articles plugin

### LINKED ISSUE #1010: Remove AMP support
Now that an [AMP plugin](https://github.com/ampproject/amp-wp) exists, having FBIA create AMP is unnecessary.

it would be good to remove this feature.

paulschreiber:
This PR:
* removes AMP support

Fixes #1010


koconder:
+1

everton-rosario:
We are ok with the removal, though we can't simply remove this before a warning of the deprecation first.
I'd say we will need another PR just deprecating this setting first and then we release a breaking change with this PR that properly removes the feature.

Not approving/merging this yet until we get the deprecation diff.

jerclarke:
At minimum, a post on the Instant Article blog announcing the change. 

A warning in the admin of anyone affect should be visible and annoying, so that they can't avoid the fact that this happened. We need to encourage them to turn on another AMP plugin, as suddenly removing all your AMP pages must be terrible from an SEO  perspective.

BalaMD:
My SiTE have a issues of amp [https://www.naijabasic.ng](https://www.naijabasic.ng)

PR #1018, opened 2019-02-26
use /ia/ as endpoint for IA markup

paulschreiber:
This PR:

* changes the IA markup endpoint from ?ia_markup=1 to /ia/ 
* allows IA markup endpoints to be cached

### DIFF
--- a/facebook-instant-articles.php
+++ b/facebook-instant-articles.php
@@ -615,7 +615,9 @@ if ( version_compare( PHP_VERSION, '5.4', '<' ) ) {
 			$post = get_post();
 			$adapter = new Instant_Articles_Post( $post );
 			$article = $adapter->to_instant_article();
-			echo $article->render( null, true );
+			$result = $article->render( null, true );
+
+			echo apply_filters( 'instant_articles_rendered_content', $result );
 
 			die();
 		}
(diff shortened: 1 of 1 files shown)


facebook-instant-articles.php
```diff
@@ -350,8 +350,7 @@ function inject_ia_markup_meta_tag() {
 		// Transform the post to an Instant Article.
 		$adapter = new Instant_Articles_Post( $post );
 		if ( $adapter->should_submit_post() ) {
-			$url = $adapter->get_canonical_url();
-			$url = add_query_arg( 'ia_markup', '1', $url );
+			$url = $adapter->get_canonical_url() . 'ia/';
```
diegoquinteiro: Hey @paulschreiber, this concatenation assumes the canonical URL ends with a "/", but that's not always the case since you can change a setting in WP to make permalinks based on query string params (or simply omit the last "/"). It'd be good to decompose the URL and make sure to add the endpoint to the right place.

diegoquinteiro: Other than this, thanks for the improvement!

paulschreiber: Do you suggest decomposing the URL using `parse_url()`?

paulschreiber: @diegoquinteiro Can you clarify?

PR #1025, opened 2019-03-26
Support for installing as composer package

roborourke:
Right now when adding this plugin via wpackagist (or in my case a fork published to packagist) it fatals because it expects the plugin to be built. This checks if the FB classes are already available.


GaryJones (approved):
✅

Issue #1053, opened 2020-08-03
Consider removing caching in Instant_Articles_Post->_get_the_content()


danielbachhuber:
The caching in `Instant_Articles_Post->_get_the_content()` seems somewhat unnecessary and makes it more difficult to develop against the plugin. It would be helpful if it was removed, or there was a filter to disable it.

danielbachhuber:
This has bitten us a few times now. Any conditional based on `is_feed( INSTANT_ARTICLES_SLUG )` is only correct part of the time, because the AJAX request to validate the FB IA transformation in the post editor can set the cache.

Another approach would be to provide some helper function that returns true when FB IA is rendering.

gisedaye:
Closing, as it looks like it was fixed on the #1058 PR

PR #1058, opened 2020-09-18
Allow post content caching to be disabled

danielbachhuber:
This PR:

* [x] Adds a `instant_articles_cache_content` filter to allow post caching content to be disabled.

Relates to #1053


GaryJones (approved):
LGTM

PR #1065, opened 2020-11-11
empty check for facebook page id

hlaporthein:
page id empty check for below error

 Notice: Trying to access array offset on value of type null in /home/john/codes/news/wp-content/plugins/fb-instant-articles/facebook-instant-articles.php on line 75

This PR:

* [x] 
* [ ] 
* [ ] 

Follows #

Relates to #

Fixes #


GaryJones:
Thanks.

This will now be covered by d56c04ae63c3dcd70993eca40c5e986890892548.

PR #1069, opened 2020-12-10
Move build to CircleCI

elomar:
While dropping PHP versions not supported by Wordpress
and adding PHP Unit to the development dependencies.


GaryJones:
The `php8` branch moves this to Github Actions, and uses an even newer version of PHP and PHPUnit.

Issue #1063, opened 2020-10-19
Too few arguments provided to get_the_excerpt hook


philipjohn:
## Steps required to reproduce the problem

1. Create and activate a small plugin that hooks into the `get_the_excerpt` hook, using code like this:
```
add_filter( 'get_the_excerpt', function( $excerpt, $post ) {
    return $excerpt;
} );
```
2. Activate and set up the FBIA plugin
3. On the site, visit the `/instant-articles` URL to look at the IA feed
4. Observe this error message:
> There has been a critical error on your website
5. Check the PHP log and you should find a PHP Fatal Error like this:
> PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function, 1 passed in /wordpress/core/5.5.1/wp-includes/class-wp-hook.php on line 287 and exactly 2 expected in /srv/htdocs/wp-content/plugins/custom-plugin.php...

## Expected Result

The IA feed should display at `/instant-articles` without error.

## Actual Result 
The error mentioned above is shown and there is a PHP Fatal Error triggered.

This is because the `get_the_excerpt` filter, [as per the documentation](https://developer.wordpress.org/reference/hooks/get_the_excerpt/), requires two parameters to be passed to it. Only one is, however, which means that any plugin correctly hooking into that filter will result in a Fatal Error.

## Version Info

* Plugin version: 4.2.1
* WordPress version: 5.5.1
* PHP version: 7.3.23

PR #1082, opened 2021-09-27
fix: pass the missing $post param to the get_the_excerpt filter

dkoo:
Fixes a bug which results in a fatal error due to a missing argument passed to the `get_the_excerpt` filter hook.

Fixes https://github.com/Automattic/facebook-instant-articles-wp/issues/1063.


yogeshbeniwal:
@dkoo Thanks for quick pull request. Looks like plugin was last updated 1 year back.

For the time being made code changes manually from this pull request to resolve the issue.

GaryJones (approved):
✅

